PR #2, opened 2017-04-05
Fix extension adding in "search_path" (Windows)

krabicezpapundeklu:
Original code checks for "program", "program.exe", "program.exe.com" and "program.exe.com.bat" instead of "program", "program.exe", "program.com" and "program.bat".
This change fixes this.


klemens-morgenstern:
Thanks!

PR #7, opened 2017-04-23
Fix line endings

Lastique:
This PR addresses https://github.com/klemens-morgenstern/boost-process/issues/88:

- Convert line endings in text files to LF.
- Add `.gitattributes` to avoid line endings issues in the future.


klemens-morgenstern:
Thanks, did it on the other repo. I'll push it here later on.

PR #10, opened 2017-05-22
Fix some typos.

triccardi-systran:


klemens-morgenstern:
Thx, my german interpretation of english is not always correct it seems.

PR #11, opened 2017-07-19
Respect WIFSIGNALED

thomasschaub:
This is making the library a little nicer. Consider this example:

    // child.cpp
    int main()
    {
      throw 0;
    }

    // parent.cpp
    #include <boost/process.hpp>
    int main()
    {
      boost::process::child("./child").wait();
    }`

In the current state, this will output the following:

     % ./parent
    libc++abi.dylib: terminating with uncaught exception of type int
    libc++abi.dylib: terminating with uncaught exception of type boost::process::process_error: waitpid(2) failed: No child processes
    zsh: abort      ./parent`

With this patch, you get this output instead

     % ./parent
    libc++abi.dylib: terminating with uncaught exception of type int
    libc++abi.dylib: terminating with uncaught exception of type boost::process::process_error: process terminated due to receipt of a signal
    zsh: abort      ./parent`

I want to point out again that this is only about making the library a little nicer, i.e., the patch is certainly not required. However, I spent a bit of time debugging my application and the error `waitpid(2) failed: No child processes` sent me down the wrong path in that process, so I think it's useful. :-)

### DIFF
--- a/include/boost/process/detail/posix/wait_for_exit.hpp
+++ b/include/boost/process/detail/posix/wait_for_exit.hpp
@@ -26,9 +26,11 @@ inline void wait(const child_handle &p, int & exit_code)
     do
     {
         ret = ::waitpid(p.pid, &status, 0);
-    } while (((ret == -1) && (errno == EINTR)) || (ret != -1 && !WIFEXITED(status)));
+    } while (((ret == -1) && (errno == EINTR)) || (ret != -1 && !WIFEXITED(status) && !WIFSIGNALED(status)));
     if (ret == -1)
         boost::process::detail::throw_last_error("waitpid(2) failed");
+    if (WIFSIGNALED(status))
+        throw process_error(std::error_code(), "process terminated due to receipt of a signal");
      exit_code = status;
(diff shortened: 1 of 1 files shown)


klemens-morgenstern:
Thanks for the PR - I have to think about that. It seems right, but then I have not concept of the whole signalling business, that I might only want to throw for terminate (i.e. `SIGKILL`). Though that could mean that I already handle the signal wrong now, and you make that nicer. I'll get back to it after I had time to think that through. Could be some time though.

egorpugin:
Why this change was only for one wait function inside that file?
Aren't other methods can be still affected by the original issue?

thomasschaub:
True, I just had a quick look, I should have changed the `wait`s in `wait_group.hpp` as well. I'll create another PR.

klemens-morgenstern:
@egorpugin @thomasschaub Please create further PRs against https://github.com/klemens-morgenstern/boost-process so the CI will test every PR.

klemens-morgenstern:
And thanks for the work!

PR #12, opened 2017-07-23
Fix incorrect on_exit() callback called on any child exited.

egorpugin:
signal_set's async_wait is only called after successful fork.
Add pid field to wait_handler.
Use waitpid instead of wait. Now the callback is called when correct child is dead. Otherwise we reset async_wait.
Add 3 tests for single ios, double ios and double threaded ios.
Tested on ubuntu 16.04.

I know it's a bit late, but can this fix to be shipped in 1.65.0?


egorpugin:
Issue for reference https://github.com/klemens-morgenstern/boost-process/issues/109

klemens-morgenstern:
Thanks man!

klemens-morgenstern:
I'll pull it into master if it works at ci https://travis-ci.org/klemens-morgenstern/boost-process

PR #13, opened 2017-07-23
Set correct access rights. Fix windows build.

egorpugin:
Accidentally set 755 flags.
Add filesystem dep to fix windows build.


egorpugin:
(Saw your CIs failed on develop branch with my changes.)
Also I see `wait.cpp` test failed on ubuntu locally. 
Maybe you could look at the test suite for known failure patterns.
It's a bit hard for me right now with one-day boost.build and little boost.process knowledge.

egorpugin:
My thoughts about that failed `double_wait` test.
https://github.com/boostorg/process/blob/develop/test/wait.cpp#L68

I think it fails, because we receive only one notification from kernel about died child. For io service in async manner, for `c.wait()` in sync manner.
From my pov using io service sets everything to async, so we should not call `c.wait()`.
... or from signal_set handler we must somehow notify child that it can continue execution from inside `c.wait()` call.

So, from design point: either test is incorrect or we must fix library.

klemens-morgenstern:
Fixed it, thanks!

PR #24, opened 2017-11-02
Fix move assignment for child_decl

ca-l-eb:
_terminate field is now copied from the move assingment operator on
child. This fixes the issue that a child process could previously only
be terminated() once (even after reassignment).


klemens-morgenstern:
My bad, thanks!

PR #34, opened 2018-03-11
Switch WINAPI calling convention macros to the replacements from Boost.WinAPI

Lastique:
WINAPI macro definition in Boost.WinAPI is deprecated as it may clash with the macro defined in Windows SDK.


Lastique:
This can be merged to master now.

PR #52, opened 2018-09-27
Changes still_active constant to support Solaris.

### LINKED ISSUE #51: Recent changes to is_running failing on Solaris
With the recent addition of the `WIFSIGNALED` macro use the `is_running` function is consistently returning false do to differences in the behavior of this macro on Solaris. 

First off the static assert fails. The code `0177` results in `WIFSIGNALED` returning `true` rather than `false`. I commented that out to see if things would run.

This is where it gets strange. After the successful execution a child the status code is `0177`, which when checked in `is_running` results in a `WIFSIGNALED` being true and the function returning false. A check of the child's exit code, using `WTERMSIG` since `WIFSIGNALED` is true, results in the value `0177` (127), which appears to not be a valid signal code on Solaris.

Looking at *wait.h* on Solaris it defines these macros quite differently than say MacOS. 
On Solaris:
```
#define WIFEXITED(stat)         ((int)((stat)&0xFF) == 0)
#define WIFSIGNALED(stat)       ((int)((stat)&0xFF) > 0 && \
                                    (int)((stat)&0xFF00) == 0)
#define WIFSTOPPED(stat)        ((int)((stat)&0xFF) == 0177 && \
                                    (int)((stat)&0xFF00) != 0)
#define WEXITSTATUS(stat)       ((int)(((stat)>>8)&0xFF))
#define WTERMSIG(stat)          ((int)((stat)&0x7F))
#define WSTOPSIG(stat)          ((int)(((stat)>>8)&0xFF))
```
From this we can clearly see why we get the values we do on Solaris when code is `0177`, since `(int)((stat)&0xFF) > 0` is true and so is `(int)((stat)&0xFF00) == 0`.

On MacOS:
```
#define _WSTATUS(x)     (_W_INT(x) & 0177)
#define _WSTOPPED       0177 
#define WSTOPSIG(x)     (_W_INT(x) >> 8)
#define WIFCONTINUED(x) (_WSTATUS(x) == _WSTOPPED && WSTOPSIG(x) == 0x13)
#define WIFSTOPPED(x)   (_WSTATUS(x) == _WSTOPPED && WSTOPSIG(x) != 0x13)
#define WIFEXITED(x)    (_WSTATUS(x) == 0)
#define WIFSIGNALED(x)  (_WSTATUS(x) != _WSTOPPED && _WSTATUS(x) != 0)
#define WTERMSIG(x)     (_WSTATUS(x))
```
Clearly we can see that the code `0177` case is handled in the `WIFSIGNALED` and would result in the expected `false`.

It almost feels like the Solaris implementation is missing the *stopped* case in it's &*signaled* check.

Not being an expert in this particular area I hesitate to make any changes other then to revert back to prior implementations that only used the `WIFEXITED` macro but would like to have an answer for how to upgrade in the future.

Any thoughts on how to proceed?

pivotal-jbarrett:
The use of the WIFSTOPPED bits did not go far enough on Solaris. It
is expected to also have a non-zero value in the next bytes. This
new value is tested be valid on Linux, MacOS and Solaris.

Fixes #51

### DIFF
--- a/include/boost/process/detail/posix/is_running.hpp
+++ b/include/boost/process/detail/posix/is_running.hpp
@@ -16,8 +16,11 @@ namespace boost { namespace process { namespace detail { namespace posix {
 // Use the "stopped" state (WIFSTOPPED) to indicate "not terminated".
 // This bit arrangement of status codes is not guaranteed by POSIX, but (according to comments in
 // the glibc <bits/waitstatus.h> header) is the same across systems in practice.
-constexpr int still_active = 0x7F;
-static_assert(!WIFEXITED(still_active) && !WIFSIGNALED(still_active), "Internal Error");
+constexpr int still_active = 0x017f;
+static_assert(WIFSTOPPED(still_active), "Expected still_active to indicate WIFSTOPPED");
+static_assert(!WIFEXITED(still_active), "Expected still_active to not indicate WIFEXITED");
+static_assert(!WIFSIGNALED(still_active), "Expected still_active to not indicate WIFSIGNALED");
+static_assert(!WIFCONTINUED(still_active), "Expected still_active to not indicate WIFCONTINUED");
 
 inline bool is_running(int code)
 {


klemens-morgenstern:
Awesome. Yeah, the exit code is garbage before the program has exited anyhow, sp the one doesn't hurt. Thanks, that's really elegantnif it solves your solatis issue.

PR #54, opened 2018-10-20
removes boost::asio:: from example in doc

tnixeu:
In an example `boost::asio::io_service ios;` is used like this and looks in the generated doc like `boost::asio::boost::asio::io_service ios;` This pull request removes `boost::asio::` from the example.

### DIFF
--- a/doc/extend.qbk
+++ b/doc/extend.qbk
@@ -5,7 +5,7 @@
 [def __on_success__ [memberref boost::process::extend::handler::on_success on_success]]
 [def __posix_executor__ [classref boost::process::extend::posix_executor ex::posix_executor]]
 [def __windows_executor__ [classref boost::process::extend::windows_executor ex::windows_executor]]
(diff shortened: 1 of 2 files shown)


tnixeu:
I fixed some more of those issues. I am not sure if I have to change the pull request or make a new one. I see both commits listed here. So I think I do not have to make a new pull request.

klemens-morgenstern:
Why would you remove `boost::asio`? I mean I should rename it to `io_context`, but I don't see why you would remove the scope.

keith-bennett-gbg:
There's a few places where the namespace prefix is duplicated such that it says `boost::asio::boost::asio::`in the rendered documentation. Which is fine enough for someone who's paying attention and knows that typos happen and can reason about it. But it's annoying anyway. I think that's partially what this is trying to resolve.

tnixeu:
I removed it, because it is added during generating the doc again. That is the first commit.
In the second commit I fixed also some other replacement issues. Please have a look at the generated doc.

klemens-morgenstern:
Oh, right, because the macro generates the prefix. Thanks, I'll update it to 'io_context' in another place.

Issue #29, opened 2018-01-19
Support BOOST_ASIO_NO_DEPRECATED


hia3:
#define BOOST_ASIO_NO_DEPRECATED 1
    #include <boost/process.hpp>
    int main() {}

Error:

    boost/process/detail/windows/async_pipe.hpp:102:39: error: no member named 'get_io_context' in 'boost::asio::windows::stream_handle'
            _sink = handle_type(_sink.get_io_context());

klemens-morgenstern:
It might be a good idea to support that, but - I'm sorry to say -  I won't get around to implement that anytime soon. I'd accept PRs though

krojew:
This would be a good addition, since boost::process now fails to build whenever this define is used.

klemens-morgenstern:
Fixed, will publish soon.

PR #67, opened 2019-01-04
Use the non-deprecated ASIO APIs.

amensel:
I believe this addresses: https://github.com/boostorg/process/issues/29

EDIT: Not-so-unrelated errors, on closer reading.


klemens-morgenstern:
Awesome, thanks for the work!

PR #74, opened 2019-03-02
Fixes to work with latest asio with custom I/O executors change

chriskohlhoff:
This pull request illustrates the minimum changes necessary for boost.process to compile correctly against the latest asio. I have only tried it on macOS, so the Windows changes may not compile as is, but they should give the idea of what needs to change.


klemens-morgenstern:
Thanks,

I already saw the changes on the other repo and fixed it.

PR #76, opened 2019-03-20
Fix signed/unsigned comparison

orgads:


klemens-morgenstern:
Thanks but I fear other tools will then complain about the C-style cast. Will add a static cast there.

orgads:
That's not C-Style, it's function-style, which is C++.

klemens-morgenstern:
You're right, but a compiler might still warn about this. I personally don't mind that for brevity sake. But you know, you'd replace one warning with another warning. If you change it, I'll merge it, otherwise I'll add it in the next days myself. Thanks!

orgads:
I don't know any compiler that warns about it. It's just like any other ctor call. But I suppose you'd want static_cast anyway for consistency, so there you have it :)

PR #113, opened 2019-12-05
Use = default for default constructor

jonesmz:


SiggyBar:
Commit `8541cae` causes this compilation error for Visual C++ 14.1 (and 14.2):

`boost\include\boost/process/child.hpp(35): error C2600: 'boost::process::child::child': cannot define a compiler-generated special member function (must be declared in the class first)`

The same can be seen here:
https://ci.appveyor.com/project/klemens-morgenstern/boost-process/branch/develop

jonesmz:
That... Seems like a compiler bug, given its declared and defined in the class only?

SiggyBar:
I created a self-contained repro case and posted it to the Visual Studio Developer Community:
https://developercommunity.visualstudio.com/content/problem/909914/unexpected-compilation-error-for-compiler-generate.html

Naios:
See #152

PR #126, opened 2020-03-08
Pass io_context in on_exit example

Martchus:
The description already states that an io_context needs to be passed. This change is adjusting the example accordingly.


Martchus:
@klemens-morgenstern I'm just pinging you because I've got not feedback so far and you might have simply forgotten the PR.

PR #128, opened 2020-03-13
Use move construction instead of overloads for const& vs &&

jonesmz:


jonesmz:
Why did you close?

klemens-morgenstern:
Moving is more expensive than passing a reference. Not that it matters much, but technically your solution would be slower.

jonesmz:
I strongly disagree.

This change is in very small functions that are implicitly inline.

I would be surprised if there was any difference in resulting assembly code, but this PR makes the code much easier to read and understand, and should be slightly faster to compile.

PR #129, opened 2020-03-13
Minor tweaks to async_pipe.hpp

jonesmz:


klemens-morgenstern:
This will cause a bunch of linker errors since you removed all the `inline` keywords.

jonesmz:
I removed the inline keyword on the in-class declaration, where the inline keyword is superfluous and redundant.

The inline keyword is preserved on the definitions of the member functions, which is the only place the keyword has an effect.

Please reconsider.

klemens-morgenstern:
You removed it from all the constructor definitions though. Did you test this?

jonesmz:
I count 3 out-of-class definitions for async_pipe::async_pipe(), and I added the inline keyword to all of them.

Is there an out-of-class definition that I missed?

klemens-morgenstern:
Ah you're right, but why did you move it to the definitions?

jonesmz:
The inline keyword is redundant when used on things inside of a class declaration.

The fact that these functions are inline (or not inline) is not directly relevant to the consumer of the declaration, because it's not part of the API.

The inline-ness is an implementation detail, so it should be only shown on the definition.

jonesmz:
My motivation here was that I was very confused when trying to understand how async_pipe worked. The inline keyword made it more difficult than it needed to be to understand.

PR #130, opened 2020-03-13
Use move construction for boost::process::detail::cmd::operator()

jonesmz:


klemens-morgenstern:
That actually introduces an additional copy.

jonesmz:
Elaborate please?

I don't see how this could introduce an additional copy. At worse its the same number.

klemens-morgenstern:
Ah you're right. You have one copy and one move or two moves. Currently you have one of either, plus passing a reference.

jonesmz:
I'm deleting this branch. You take too long to do anything, and you close PRs for reasons that you later admit are wrong, but then leave the PR closed instead of doing anything about being wrong.

PR #131, opened 2020-03-13
Use move construction for start_dir

jonesmz:


klemens-morgenstern:
Unneeded, `api::start_dir_init` copies anyhow.

jonesmz:
The objective of the patches that add the use of std::move is to bring the API for boost::process to a state of api consistency, for most types. That would result in types that didn't use the pass-by-value-and-std-move idiom to be easily identifiable as unusual and worth additional attention.

PR #132, opened 2020-03-13
use = default for struct shell_

jonesmz:


klemens-morgenstern:
Rejecting for MSVC 14, see #152

PR #133, opened 2020-03-13
Make constructor for codecvt_category_t constexpr

jonesmz:


klemens-morgenstern:
Doesn't work for C++11

jonesmz:
Why are you supporting a 9 year old language?

klemens-morgenstern:
LTS

jonesmz:
Ok. I suppose that's not a big problem for me.

Can you leave this patch open, and tag it with some kind of "Requires C++14" tag?

Unless you don't want the change for other reasons, of course.

klemens-morgenstern:
It is rather unlikely that I will build a C++14 version of this at any point. A C++20 boost.process2 is much more likely, i.e. a reference implementation of P1750R1.

jonesmz:
Ok.

PR #134, opened 2020-03-13
Minor tweaks to error.hpp

jonesmz:


jonesmz:
There are a lot of other improvements that this PR included beyond only the one function change.

klemens-morgenstern:
Which beside whitespace and replacing `typedef` with `using`?

jonesmz:
Whitespace and `using` are valuable on their own merits.

But to directly answer your question, minus those:     void on_error(Executor&, const std::error_code & ec) const noexcept(false)

klemens-morgenstern:
Right, I saw that, what's the benefit? Isn"t that the exact same thing as having the on `noexcept` specifier?

jonesmz:
noexcept(false) is explicitly saying "This function throws an exception".

This is useful for static analysis tools, IDEs that do IntelliSense / code completion, and just in general makes it very clear to the end-user reading the function signature that the function is expected to throw.

klemens-morgenstern:
It"s in the detail namespace, no library consumer should ever see that part of the code. 
I am gonna keep it to the `noexcept` I added in my commit.

PR #135, opened 2020-03-13
Use move construction for the std::string in boost::process::detail::exe_::operator()

jonesmz:


klemens-morgenstern:
Template is for forward-declaration & your changes don't avoid any copy.

jonesmz:
Your statement about the template being for forward declaration does not make any sense. Please elaborate.


This change. By itself. Does not avoid copying. But it does enable other code to avoid copying by using std::move when calling the functions modified in this commit.

The trade off for this change is either copy sizeof(void*) and then dereference, or copy 2*sizeof(void*) without a dereference. 

However the status quo makes it impossible for users of these functions to std::move() into these functions, whereas this commit enables that as a future improvement.

klemens-morgenstern:
If `boost::filesystem::path` is a forward declaration and the function are not templated, the compiler will try to compile them. if it's a template with a default argument the compiler will be compile them on usage, at which point `filesystem::path` would be available.

Adding lvalue reference overloads would be the right way to do that, or rather just `string_view`'s if I ever chose to make a breaking change.

PR #136, opened 2020-03-13
Fix "comparison between signed and unsigned" warning

jonesmz:


jonesmz:
Something is wrong with the unit tests on macosx.

This change should have had no effect on, for example,

bind_stdout.cpp:131: error: in "bind_stdout/nul": check 0 == exit_code has failed [0 != 11]


But it's failing.

PR #141, opened 2020-03-13
Fixes #122

### LINKED ISSUE #122: Incorrect std::equal usage triggers UB, detected by address sanitizer
In environment.hpp, there are several std::equal calls like this:

    std::equal(st1.begin(), st1.end(), *p)

`st1` is a string, with user specified content. `p` is a pointer to a character (I surmise, as `auto` is used all over the place), and holds an existing key of `*this` in each iteration.
This usage of equal is incorrect, as it yields to reading invalid data if `size(st1) > strlen(p)`.

Address sanitizer (ASAN) catches the issue. To reproduce:

```
bp::environment e = boost::this_process:environment();
e["foobarbazquxmux"] = "xyz";
```

The issue can be worked around by using the `set` member of the implementation directly:

```
/*bp::environment*/ auto e = boost::this_process:environment();
e.set("foobarbazquxmux", "xyz");
```

jonesmz:


jonesmz:
This issue is causing me pain.

Even if the ASAN output is a false positive, it is still triggering on every ASAN run in my test environment. This produces a lot of undesired log output.

How can we improve the boost process code to ensure this does not trigger ASAN output anymore?

PR #144, opened 2020-03-13
Change windows pid_t to decltype(::boost::winapi::PROCESS_INFORMATION_::dwProcessId);

jonesmz:


jonesmz:
Your commit misses half of the improvements .

Please reconsider.

klemens-morgenstern:
Which? What's the improvement of `using` over `typedef`?

jonesmz:
The use of decltype(::boost::winapi::PROCESS_INFORMATION_::dwProcessId);

and the *removal* of the static cast. Keeping the static cast will result in changes to the type of dwProcessId going unaccounted for because the static cast masks it.

klemens-morgenstern:
`dwProcessId` literally says it's a `DWORD`. No need for the decltype.

jonesmz:
Until dwProcessId changes...

klemens-morgenstern:
Google hungarian notation.

jonesmz:
I don't see how that's relevant.

Just because hungarian notation was used when the variable was originally defined, does not in any way imply that it's type won't change in the future. In fact, it's likely that the name of the variable would be left alone to prevent causing compile errors in other people's code even if the type of it changes.

You can either assume that no one will ever change the type of dwProcessId, or you can be guaranteed that your code either:
1) Stops compiling, if dwProcessId has it's name changed
2) Continues working perfectly if dwProcessId has it's type changed but not it's name.

In either situation, using decltype is better than having the type changed out from under you without the type of pid_t being updated to match.

klemens-morgenstern:
Yes, and if the WinApi changes the type (which I would consider very unlikely) `::boost::winapi::PROCESS_INFORMATION_::dwProcessId` will need to be changed, too.

jonesmz:
Right.

So using decltype(::boost::winapi::PROCESS_INFORMATION_::dwProcessId)

means that boost::process will not need to be updated, unless the name of dwProcessId changes.

PR #145, opened 2020-03-13
Fixes #188

jonesmz:


jonesmz:
Removing the inline documentation from my commit is not a great choice.

It took me a while to get all that information.

PR #152, opened 2020-03-23
[Build Fix] Workaround for a MSVC 14.1 and 14.2 compiler bug

Naios:
Currently affects the MSVC CI

* boost\include\boost/process/child.hpp(35): error C2600:
    'boost::process::child::child': cannot define a compiler-generated
    special member function (must be declared in the class first)
* Introduced by 8541cae
* See boostorg/process#113

### DIFF
--- a/include/boost/process/detail/child_decl.hpp
+++ b/include/boost/process/detail/child_decl.hpp
@@ -73,7 +73,7 @@ public:
 
     template<typename ...Args>
     explicit child(Args&&...args);
-    child() = default;
+    child() { } // Must be kept non defaulted for MSVC 14.1 & 14.2 #113
     child& operator=(const child&) = delete;
     child& operator=(child && lhs)
     {


jonesmz:
Which actual Visual Studio release does this correspond to?

jonesmz:
https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B#Internal_version_numbering

MSVC++ 14.2  _MSC_VER == 1920 (Visual Studio 2019 Version 16.0)


Has this bug been posted to Microsoft's community bug tracker?

jonesmz:
Well, I can't stop this change from being merged, but my opinion is that this is a compiler bug. Boost::Process shouldn't work around compiler bugs.

Naios:
I know it seems not like an optimal solution to adapt the library to a compiler bug however you also don't want to lock out the majority of the userbase from a fairly new compiler (~3 years old) because of a single line which doesn't introduce new functionality or bugfixes.

klemens-morgenstern:
I originally changed most of the constructors to work in the same way, for exactly that compiler.  Which is why many initializers have constructors like that. It's valid C++ and I do understand that not everone can update their compilers as often.

jonesmz:
For me, the big sticking point here is that this isn't an issue of people being able to upgrade their compilers or not. This is an issue of Microsoft having a non-conforming compiler, for it's latest release.

By adapting the code to work with Microsoft's latest bugs, you're implicitly saying that's acceptable behavior from Microsoft, and drastically reducing the number of people that would put pressure on them to fix the problem.

PR #158, opened 2020-05-14
Add the lambda declarator needed to disambiguate the multiple return …

acronce:
…values from these lambda functions as a bool. Before this change the return of a "false" was interpreted as an int, but the second return is a bool. This caused the following compile error when accessing environment variables under Xcode 11.4.1: Return type 'bool' must match previous return type 'int' when lambda expression has unspecified explicit return type

### DIFF
--- a/include/boost/process/detail/posix/environment.hpp
+++ b/include/boost/process/detail/posix/environment.hpp
@@ -230,7 +230,7 @@ template<typename Char>
 inline auto basic_environment_impl<Char>::get(const string_type &id) -> string_type
 {
     auto itr = std::find_if(_data.begin(), _data.end(), 
-            [&](const string_type & st)
+            [&](const string_type & st) -> bool
             {
                 if (st.size() <= id.size())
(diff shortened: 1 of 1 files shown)


jonesmz:
This is a straight forward bug in XCode.

There was never any ambiguity in the lambdas in question. Their compiler assuming that "false" is an int, and then erroring on the std::equals()... line is not something that should be worked around.

klemens-morgenstern:
It is something that needs a workaround if someone wants to use `XCode` & `boost.process` together.

acronce:
Thanks for accepting the pull request. 

I agree that it's irritating to need this work around, but an explicit return type is not a bad thing.

FYI, Xcode uses clang under the hood, so this really is a clang issue. Apple builds their own copy of clang with their own version number (not very helpful). 

The version returned by clang is "Apple clang version 11.0.3 (clang-1103.0.32.59)". According to wikipedia that maps to LLVM 9.0.0.

PR #181, opened 2020-11-04
Fix cross-compile linux->mingw

orgads:
Headers are lower case.


Martchus:
I've just ran into this problem as well. It would be great if the patch could be merged. I can confirm that it works.

xantares:
+1 cc @klemens-morgenstern

orgads:
Rebased.

klemens-morgenstern:
Has this been tested on windows for native compiles?

orgads:
Yes

xantares:
on native windows the filesystem is case-insensitive so either works but not on linux

PR #191, opened 2020-12-12
posix/executor: Fix uninitialized memory access in case of child process error

dkl:
The parent process didn't validate the amount of bytes received from the child process, so it sometimes accessed uninitialized memory. This could be triggered by error cases using the old write_error() function, which sends 4 + 4 + N bytes instead of 8 + N as assumed by _read_error().

For example: Giving an invalid file name for stdout/stdin redirection (or if the file cannot be opened for some other reason):
```
std::string empty_filename;
bp::child c(argv, bp::std_out > empty_filename);
```

### DIFF
--- a/include/boost/process/detail/posix/executor.hpp
+++ b/include/boost/process/detail/posix/executor.hpp
@@ -151,13 +151,7 @@ class executor
 
     void write_error(const std::error_code & ec, const char * msg)
     {
-        //I am the child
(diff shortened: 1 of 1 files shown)


klemens-morgenstern:
I don't understand why this would cause an error? Wouldn't the buffer of the pipe avoid that?

Issue #196, opened 2021-01-21
async_pipe does not meet AsyncReadStream requirements


gulachek:
The async_pipe class does not meet the requirements for an [AsyncReadStream](https://www.boost.org/doc/libs/1_75_0/doc/html/boost_asio/reference/AsyncReadStream.html).  The get_executor() member should be implemented, probably to just return the executor of the underlying implementation.  The impact is that async operations following the guidelines of posting work on the associated executor of the operation will not compile since this member is not available.

olk:
I pushed pull request #197 in order to fix this issue.

gulachek:
I don't understand how using post() instead of dispatch() fulfills the request to add a get_executor() member function. 197 doesn't fix this.

PR #197, opened 2021-01-22
use post() instead dispatch() for async_await()

olk:
- dispatch() might queue the completion handler or invoke it immediately
- this might result in problems if this code runs in  completion
  handlers like coroutines where you must resume an already running coroutine
- using post() solves this issue


olk:
relates to issue  #196

gulachek:
This might fix another issue you found, but this documented issue (196) isn't fixed by the diff I'm seeing.  If you see the AsyncReadStream requirements, the async_pipe needs to implement a get_executor member function.

olk:
This patcj drom above seams to solve our production code.

klemens-morgenstern:
I think the outer one should be dispatch, but the inner one should post.

PR #236, opened 2022-01-18
Use bp::extend::{on_setup/on_error} in windows example

achow101:
Otherwise it does not compile.


fanquake:

klemens-morgenstern:
Fixed already.

PR #237, opened 2022-01-21
Avoid using boost::filesystem w/string conversions on Windows

achow101:
Converting from wstring to string (or vice versa) with boost::filesystem requires Filesystem to be compiled. However if these conversions are avoided, then boost::filesystem will not be needed by downstream projects.

A project I am working on is trying to remove boost filesystem and we noticed that doing so results in a linker error when cross compiling to windows. The root cause was identified to be that boost process for Windows ends up using some conversion functions which require boost filesystem to be compiled. This patch resolved those issues for us.


fanquake:

fanquake:
Can this be closed after #264?

Issue #183, opened 2020-11-11
Missing include in windows/handles.hpp


Tradias:
Compiling code that uses `boost::process::child` with a custom `on_setup` handler with MSVC 16.8 throws the following errors (boost 1.74):

```
[build] boost/process/detail/windows/handles.hpp(142): error C2039: "GetHandleInformation" ist kein Member von "boost::winapi".
[build] boost/process/detail/windows/handles.hpp(142): note: Diese Diagnose trat in der vom Compiler generierten Funktion "void boost::process::detail::windows::limit_handles_::on_setup(Executor &) const" auf.
[build] boost/process/detail/windows/handles.hpp(143): error C2039: "HANDLE_FLAG_INHERIT_" ist kein Member von "boost::winapi".
[build] boost/winapi/dll.hpp(190): note: Siehe Deklaration von "boost::winapi"
[build] boost/process/detail/windows/handles.hpp(143): note: Diese Diagnose trat in der vom Compiler generierten Funktion "void boost::process::detail::windows::limit_handles_::on_setup(Executor &) const" auf.
[build] boost/process/detail/windows/handles.hpp(154): error C2039: "SetHandleInformation" ist kein Member von "boost::winapi".
[build] boost/winapi/dll.hpp(190): note: Siehe Deklaration von "boost::winapi"
```

The missing functions are declared in `#include <boost/winapi/handle_info.hpp>`. Adding that include to `boost/process/detail/windows/handles.hpp` solves the problem.

Tradias:
Remains to be an issue in Boost 1.77 and MSVC 16.11

oxygene:
Still a problem with msvc143 (Visual Studio 2022, v19.30) and boost 1.78.0.

```
C:\dev\boost-1.78.0\boost-1_78\boost/process/detail/windows/handles.hpp(141): error C2039: 'INVALID_HANDLE_VALUE_': is not a member of 'boost::winapi'
C:\dev\boost-1.78.0\boost-1_78\boost/winapi/get_current_process_id.hpp(26): note: see declaration of 'boost::winapi'
C:\dev\boost-1.78.0\boost-1_78\boost/process/detail/windows/handles.hpp(132): note: This diagnostic occurred in the compiler generated function 'void boost::process::detail::windows::limit_handles_::on_setup(Executor &) const'
[...]

Issue #224, opened 2021-10-12
Compilation fails with '/permissive-' in Boost 1.75.0 when "boost/process/group.hpp" is included


vaishnavkatiyar:
The following error is generated when the command `cl /I.\.. /EHsc /permissive- Source.cpp` is used:
```
Microsoft (R) C/C++ Optimizing Compiler Version 19.27.29112 for x86
Copyright (C) Microsoft Corporation.  All rights reserved.

Source.cpp
.\..\boost/process/detail/windows/handles.hpp(146): error C2039: 'GetHandleInformation': is not a member of 'boost::winapi'
.\..\boost/winapi/dll.hpp(190): note: see declaration of 'boost::winapi'
.\..\boost/process/detail/windows/handles.hpp(146): note: This diagnostic occurred in the compiler generated function 'void boost::process::detail::windows::limit_handles_::on_setup(Executor &) const'
.\..\boost/process/detail/windows/handles.hpp(147): error C2039: 'HANDLE_FLAG_INHERIT_': is not a member of 'boost::winapi'
.\..\boost/winapi/dll.hpp(190): note: see declaration of 'boost::winapi'
.\..\boost/process/detail/windows/handles.hpp(147): note: This diagnostic occurred in the compiler generated function 'void boost::process::detail::windows::limit_handles_::on_setup(Executor &) const'
.\..\boost/process/detail/windows/handles.hpp(158): error C2039: 'SetHandleInformation': is not a member of 'boost::winapi'
.\..\boost/winapi/dll.hpp(190): note: see declaration of 'boost::winapi'
.\..\boost/process/detail/windows/handles.hpp(158): note: This diagnostic occurred in the compiler generated function 'void boost::process::detail::windows::limit_handles_::on_setup(Executor &) const'
.\..\boost/process/detail/windows/handles.hpp(158): error C2039: 'HANDLE_FLAG_INHERIT_': is not a member of 'boost::winapi'
.\..\boost/winapi/dll.hpp(190): note: see declaration of 'boost::winapi'
.\..\boost/process/detail/windows/handles.hpp(158): note: This diagnostic occurred in the compiler generated function 'void boost::process::detail::windows::limit_handles_::on_setup(Executor &) const'
.\..\boost/process/detail/windows/handles.hpp(165): error C2039: 'SetHandleInformation': is not a member of 'boost::winapi'
.\..\boost/winapi/dll.hpp(190): note: see declaration of 'boost::winapi'
[...]

PR #240, opened 2022-02-26
fix: "GetHandleInformation" is not a member of "boost::winapi"

Tradias:
Fixes https://github.com/boostorg/process/issues/183 and https://github.com/boostorg/process/issues/224


fanquake:

klemens-morgenstern:
Seems I've already fixed this with another include.

PR #250, opened 2022-05-19
Drone

klemens-morgenstern:

PR #253, opened 2022-05-23
Remove superflous calls in CMakeLists

Flamefire:
The find_package is not required, as the dependencies are done using the superproject build
The include_directories is already there: target_include_directories

Partial revert of 610b337fa3bc82ecd19cd7baa6e90874f62c050f in #250


Flamefire:
@klemens-morgenstern Please merge soon as this blocks the Boost CMake build

I suggest to add one for this lib (see boost-ci)

Issue #263, opened 2022-08-01
1.80.0 beta 1 fails to compile for MinGW-w64 using `BOOST_PROCESS_USE_STD_FS`


hebasto:
A code example from https://www.boost.org/doc/libs/1_80_0_beta1/doc/html/process.html#boost_process.introduction fails to compile:
```
$ x86_64-w64-mingw32-g++-posix --version
x86_64-w64-mingw32-g++-posix (GCC) 10-posix 20220113
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

$ x86_64-w64-mingw32-g++-posix -std=c++17 process_demo.cpp -o process_demo.exe -I/home/hebasto/Boost.Dev/boost_1_80_0_b1/boost_1_80_0 -DBOOST_PROCESS_USE_STD_FS -lws2_32 -static
In file included from /home/hebasto/Boost.Dev/boost_1_80_0_b1/boost_1_80_0/boost/process/detail/basic_cmd.hpp:17,
                 from /home/hebasto/Boost.Dev/boost_1_80_0_b1/boost_1_80_0/boost/process/args.hpp:33,
                 from /home/hebasto/Boost.Dev/boost_1_80_0_b1/boost_1_80_0/boost/process.hpp:22,
                 from process_demo.cpp:3:
/home/hebasto/Boost.Dev/boost_1_80_0_b1/boost_1_80_0/boost/process/detail/windows/basic_cmd.hpp: In static member function ‘static std::string boost::process::detail::windows::exe_cmd_init<Char>::get_shell(char)’:
/home/hebasto/Boost.Dev/boost_1_80_0_b1/boost_1_80_0/boost/process/detail/windows/basic_cmd.hpp:162:77: error: no matching function for call to ‘std::filesystem::__cxx11::path::string(const codecvt_type&)’
  162 |     static std:: string get_shell(char)    {return shell(). string(codecvt()); }
      |                                                                             ^
In file included from /usr/lib/gcc/x86_64-w64-mingw32/10-posix/include/c++/filesystem:45,
                 from /home/hebasto/Boost.Dev/boost_1_80_0_b1/boost_1_80_0/boost/process/filesystem.hpp:9,
[...]

PR #264, opened 2022-08-02
Fix compiling for MinGW-w64 using std::filesystem

hebasto:
Fixes boostorg/process#263.


klemens-morgenstern:
Looks to be an asio issue.

PR #265, opened 2022-08-12
Add waitpid call on child error to avoid zombies

enrico-samknows:
When using process, if execve fails due to non-existing program, I see zombies process appearing.
With this patch I don't see zombies anymore.


klemens-morgenstern:
I'll look into the problem, but that doesn't look like a proper solution. It should capture the error.

enrico-samknows:
@klemens-morgenstern thanks for the reply. Feel free to fix it so it becomes compliant, or please send me more information about how you'd fix it, so I can try to improve it.

PR #281, opened 2022-11-30
Add Executable PathName From Process Identifier Implementation.

time-killer-games:
**exe_path(pid) documentation:**

`boost::filesystem::path boost::process::v2::ext::exe_path(boost::process::pid_type pid);` returns the executable pathname associated with the given process identifier in the pid argument. boost::filesystem::path::string::empty() is true if the function has failed. The return value will be an absolute pathname with all symbollic links resolved. The pathname will be normalized, meaning there will be no double slashes, no dot, and no dot-dot in the returned pathname. The returned pathname is limited to MAX_PATH characters on Windows, and on Unix-likes, the returned pathname is limited to PATH_MAX characters. PATH_MAX is implementation-defined by the platform. On OpenBSD, the platform relies on retrieving argv[0] from the assocaited process's command line arguments, to resolve the executable pathname, because OpenBSD does not have an official or dedicated API for such. The argv[0] suffix is appended to the $PWD environment variable of the associated process as the prefix, when the argv[0] suffix is a relative pathname, with the current working directory of the assocaited process as the fallback prefix. On failure, the executable pathname will have a wrong ino_t and/or dev_t value, and the function will return an empty string.


include/boost/process/v2/detail/impl/process_handle_windows.ipp
```diff
@@ -113,6 +123,33 @@ void check_running_(HANDLE handle, error_code & ec, DWORD & exit_status)
...
+  if (nt_err > 0xC0000000)
+    ec = detail::get_last_error();
+}
+
+void resume_(HANDLE handle, error_code & ec)
+{
+
+  auto nt_err = NtResumeProcess(handle);
+  if (nt_err > 0xC0000000)
+    ec = detail::get_last_error();
```
klemens-morgenstern, line 139: I'll need to check if that is actually correct.

include/boost/process/v2/detail/process_handle_fd.hpp
```diff
@@ -182,7 +182,42 @@ struct basic_process_handle_fd
...
+        resume(ec);
+        if (ec)
+            detail::throw_error(ec, "resume");
+    }
+
+    void resume(error_code &ec)
+    {
+        if (pid_ <= 0)
+            return ;
+        if (::kill(pid_, SIGTERM) == -1)
```
klemens-morgenstern, line 218: my typo?

include/boost/process/v2/detail/process_handle_fd_or_signal.hpp
```diff
@@ -211,6 +211,42 @@ struct basic_process_handle_fd_or_signal
...
+        resume(ec);
+        if (ec)
+            detail::throw_error(ec, "resume");
+    }
+
+    void resume(error_code &ec)
+    {
+        if (pid_ <= 0)
+            return ;
+        if (::kill(pid_, SIGTERM) == -1)
```
klemens-morgenstern, line 247: sam here - looks like it should be `SIGCONT`.

include/boost/process/v2/exe.hpp
```diff
@@ -0,0 +1,38 @@
...
+boost::filesystem::path exe_path(pid_type pid, error_code & ec);
+#elif defined(__NetBSD__)
+boost::filesystem::path exe_path(pid_type pid, error_code & ec);
+#elif defined(__OpenBSD__)
+boost::filesystem::path exe_path(pid_type pid, error_code & ec);
+#elif defined(__sun)
+boost::filesystem::path exe_path(pid_type pid, error_code & ec);
+#else
+#error "Platform not supported"
+#endif
```
klemens-morgenstern, line 30: all the same decl, no need for the #ifdefs.

include/boost/process/v2/impl/exe.ipp
```diff
@@ -0,0 +1,313 @@
...
+BOOST_PROCESS_V2_BEGIN_NAMESPACE
+
+namespace ext {
+
+#if defined(BOOST_PROCESS_V2_WINDOWS)
+
+boost::filesystem::path exe_path(pid_type pid, error_code & ec)
+{
+    boost::filesystem::path path;
+    if (pid == GetCurrentProcessId())
```
klemens-morgenstern, line 48: I wonder if we (1) need this and (2) if that maybe should be it's own function.

include/boost/process/v2/impl/exe.ipp
```diff
@@ -0,0 +1,313 @@
...
+            }
+        }
+    } 
+    else 
+    {
+        HANDLE proc = detail::ext::open_process_with_debug_privilege(pid);
+        if (proc == nullptr) return path;
+        wchar_t buffer[MAX_PATH];
+        DWORD size = sizeof(buffer);
+        if (QueryFullProcessImageNameW(proc, 0, buffer, &size) != 0)
```
klemens-morgenstern, line 66: Error handling?

include/boost/process/v2/impl/exe.ipp
```diff
@@ -0,0 +1,313 @@
...
+        ec = detail::get_last_error();
+    }
+    return path;
+}
+
+#elif defined(__FreeBSD__) || defined(__DragonFly__)
+
+boost::filesystem::path exe_path(pid_type pid, error_code & ec) 
+{
+    std::boost::filesystem::path;
```
klemens-morgenstern, line 125: can I get a few comments of what this is?

include/boost/process/v2/impl/exe.ipp
```diff
@@ -0,0 +1,313 @@
...
+                kinfo_file *kif = nullptr;
+                kd = kvm_openfiles(nullptr, nullptr, nullptr, KVM_NO_FILES, nullptr);
+                if (!kd) return false;
+                if ((kif = kvm_getfiles(kd, KERN_FILE_BYPID, pid, sizeof(struct kinfo_file), &cntp))) 
+                {
+                    for (int i = 0; i < cntp; i++) 
+                    {
+                        if (kif[i].fd_fd == KERN_FILE_TEXT) 
+                        {
+                            if (st.st_dev == (dev_t)kif[i].va_fsid || st.st_ino == (ino_t)kif[i].va_fileid)
```
klemens-morgenstern, line 210: that's one `if` statement.

`static_cast` plz

include/boost/process/v2/impl/exe.ipp
```diff
@@ -0,0 +1,313 @@
...
+
+boost::filesystem::path exe_path(pid_type pid, error_code & ec)
+{
+    std::boost::filesystem::path;
+    auto is_executable = [](pid_type pid, std::string in, std::string *out) 
+    {
+        *out = "";
+        bool success = false;
+        struct stat st;
+        if (!stat(in.c_str(), &st) && (st.st_mode & S_IXUSR) && (st.st_mode & S_IFREG))
```
klemens-morgenstern, line 195: can be one `if` statement.

include/boost/process/v2/impl/pid.ipp
```diff
@@ -22,6 +66,215 @@ pid_type current_pid() {return ::GetCurrentProcessId();}
 pid_type current_pid() {return ::getpid();}
 #endif
 
+#if defined(BOOST_PROCESS_V2_WINDOWS)
+
+std::vector<pid_type> all_pids(error_code & ec)
```
klemens-morgenstern, line 71: I wonder if that should be able to take a custom allocator.

test/v2/pid.cpp
```diff
@@ -12,4 +12,8 @@ BOOST_AUTO_TEST_CASE(test_pid)
 {
     namespace bp2 = boost::process::v2;
     BOOST_CHECK_NE(bp2::current_pid(), static_cast<bp2::pid_type>(0));
+
+    auto all = bp2::all_pids();
+    auto itr = std::find(all.begin(), all.end(), bp2::current_pid());
```
klemens-morgenstern, line 17: This should have two more tests: one for a child process and one for a grand child.

klemens-morgenstern (changes requested):

include/boost/process/v2/detail/impl/exe.ipp
```diff
@@ -0,0 +1,60 @@
...
+        CloseHandle(hToken);
+    }
+    return proc;
+}
+
+BOOST_PROCESS_V2_DECL
+BOOL is_x86_process(HANDLE proc) {
+    BOOL isWow = true;
+    SYSTEM_INFO systemInfo;
+    GetNativeSystemInfo(&systemInfo);
```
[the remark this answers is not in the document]
time-killer-games: That is correct, unless I missed something: https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getnativesysteminfo

include/boost/process/v2/detail/impl/exe.ipp
```diff
@@ -0,0 +1,60 @@
...
+
+#if defined(_WIN32)
+
+BOOST_PROCESS_V2_DECL
+HANDLE open_process_with_debug_privilege(pid_type pid) {
+    HANDLE proc = nullptr;
+    HANDLE hToken = nullptr;
+    LUID luid;
+    TOKEN_PRIVILEGES tkp;
+    if (OpenProcessToken(GetCurrentProcess(), TOKEN_ADJUST_PRIVILEGES | TOKEN_QUERY, &hToken)) {
```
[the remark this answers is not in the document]
time-killer-games: The error gets recorded to GetLastError(). Although the website doesn't seem to say why it would ever error, so I'm not sure what the failure cases would be without debugging it hands on.

include/boost/process/v2/detail/process_handle_fd.hpp
time-killer-games: Well it appears that was you :P. Glad to know even the best of us can typo.

time-killer-games:
Yeah... I'm going to [skip ci] for now on....

include/boost/process/v2/impl/exe.ipp
time-killer-games: that was a typo, isn't it just without the std::?

include/boost/process/v2/impl/exe.ipp
time-killer-games: you mean combine the fd if statement with dev/ino or split dev/ino into two different if statements?

test/v2/pid.cpp
time-killer-games: I won't be able to do that until I add my parent and child from pid functions. This pr is meant just for getting the exe path from a pid. If you are ok with me conflating the two pull requests I can do that if you want.

time-killer-games: In the meantime, I'm going to create a test for getting exe from path out of all pids.

include/boost/process/v2/impl/exe.ipp
time-killer-games: I explained already, we do need it. MS docs explain Only use the former function for the current process, and it is better to only get foreign process exe names with the latter otherwise you might end up with wrong results on occasion. I'll link to the documentation page which claims that on msdn if you can't find it. I will admit it does seem a bit odd.

include/boost/process/v2/impl/exe.ipp
time-killer-games: Are you saying you want me to store the value of GetLastErrorr somewhere? I thought std::error_code did this automatically with the function I am using when it returns an empty string.

include/boost/process/v2/impl/pid.ipp
time-killer-games: You mean using malloc or new instead of std::vector?

include/boost/process/v2/detail/impl/exe.ipp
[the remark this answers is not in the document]
time-killer-games: The function below it can fail, though, so i added a check.  I'm not sure what to do when that function fails, because if it returns a true, but the target process is actually x64 then it will likely segfault when reading proc info depending on the function

time-killer-games:
@klemens-morgenstern i marked a few conversations as resolved, feel free to unresolve them if i did so incorrectly.

include/boost/process/v2/impl/exe.ipp
time-killer-games: > To retrieve the module name of the current process, use the [GetModuleFileName](https://learn.microsoft.com/en-us/windows/desktop/api/libloaderapi/nf-libloaderapi-getmodulefilenamea) function with a NULL module handle. This is more efficient than calling the GetProcessImageFileName function with a handle to the current process.
> 
> To retrieve the name of the main executable module for a remote process in win32 path format, use the [QueryFullProcessImageName](https://learn.microsoft.com/en-us/windows/desktop/api/winbase/nf-winbase-queryfullprocessimagenamea) function.

https://learn.microsoft.com/en-us/windows/win32/api/psapi/nf-psapi-getprocessimagefilenamea

include/boost/process/v2/detail/impl/exe.ipp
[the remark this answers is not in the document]
time-killer-games: I think i know what you want now, you want me to add an error code argument to the helper functions so i can pass the errors on to the functions which call them. makes sense now. Taking care of it.

include/boost/process/v2/detail/impl/process_handle_windows.ipp
time-killer-games: This doesnt seem correct according to NTSTATUS documentation but if NTSATUS is nonzero assume error. Very rarely will it be a function that succeeded properly with a nonzero value. At best it would be informational, which still means something that needs the developers attention is going on.

include/boost/process/v2/detail/exe.hpp
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) 2022 Klemens D. Morgenstern
```
klemens-morgenstern, line 1: Thy name goes here.

include/boost/process/v2/detail/impl/exe.ipp
```diff
@@ -0,0 +1,71 @@
...
+
+namespace detail
+{
+
+namespace ext
+{
+
+#if defined(_WIN32)
+
+HANDLE open_process_with_debug_privilege(pid_type pid, error_code & ec) {
```
klemens-morgenstern, line 23:
```suggestion
HANDLE open_process_with_debug_privilege(pid_type pid, error_code & ec) 
{
```

Personal preference.

klemens-morgenstern: I'd also like to get a few lines on what this does and why we need it. Not to mention that `basic_process`.

include/boost/process/v2/detail/impl/exe.ipp
```diff
@@ -0,0 +1,71 @@
...
+        CloseHandle(hToken);
+    }
+    if (!proc)
+        proc = OpenProcess(PROCESS_ALL_ACCESS, false, pid);
+    if (!proc)
+        ec = detail::get_last_error();
+    return proc;
+}
+
+BOOL is_x86_process(HANDLE proc, error_code & ec) {
```
klemens-morgenstern, line 49: Just out of curiosity: Is there something like that on linux?

include/boost/process/v2/detail/impl/process_handle_windows.ipp
```diff
@@ -113,6 +123,31 @@ void check_running_(HANDLE handle, error_code & ec, DWORD & exit_status)
...
+    auto nt_err = NtSuspendProcess(handle);
+    if (nt_err > 0xC0000000)
+       ec = detail::get_last_error();
+}
+
+void resume_(HANDLE handle, error_code & ec)
+{
+    auto nt_err = NtResumeProcess(handle);
+    if (nt_err > 0xC0000000)
+        ec = detail::get_last_error();
```
klemens-morgenstern, line 138: We need to check if the GetLastError call is correct here. It's possble we need another error category, if the codes don't align.

include/boost/process/v2/impl/exe.ipp
```diff
@@ -0,0 +1,324 @@
...
+
+filesystem::path exe_path(boost::process::v2::pid_type pid, std::error_code & ec)
+{
+    std::string path;
+    char exe[PATH_MAX];
+    if (realpath(("/proc/" + std::to_string(pid) + "/exe").c_str(), exe)) 
+    {
+        path = exe;
+    }
+    if (path.empty())
```
klemens-morgenstern, line 122: This error handling probably works, but is incorrect. Doc states

> Otherwise, it returns NULL, the contents of the array  resolved_path are undefined, and [errno](https://man7.org/linux/man-pages/man3/errno.3.html) is set to indicate the error.

So the way I'd write this is as so:

```cpp
    char exe[PATH_MAX];
    if (realpath(("/proc/" + std::to_string(pid) + "/exe").c_str(), exe) != nullptr) 
        path = exe;
    else
        ec = detail::get_last_error();
```

include/boost/process/v2/exe.hpp
```diff
@@ -0,0 +1,32 @@
...
+namespace ext {
+
+// return executable path from pid
+#ifdef BOOST_PROCESS_USE_STD_FS
+BOOST_PROCESS_V2_DECL std::filesystem::path exe_path(boost::process::v2::pid_type pid, std::error_code & ec);
+BOOST_PROCESS_V2_DECL std::filesystem::path exe_path(boost::process::v2::pid_type pid);
+#else
+BOOST_PROCESS_V2_DECL boost::filesystem::path exe_path(boost::process::v2::pid_type pid, std::error_code & ec);
+BOOST_PROCESS_V2_DECL boost::filesystem::path exe_path(boost::process::v2::pid_type pid);
+#endif
```
klemens-morgenstern, line 24: I am wondering about the naming - wouldn't `executable` be a fit too?

```cpp
executable(proc.id());
```

Looks pretty good to me.

include/boost/process/v2/impl/exe.ipp
```diff
@@ -0,0 +1,324 @@
...
+        ec = detail::get_last_error();
+    return path;
+}
+
+#elif defined(__FreeBSD__) || defined(__DragonFly__)
+
+filesystem::path exe_path(boost::process::v2::pid_type pid, std::error_code & ec) 
+{
+    std::string path;
+    int mib[4];
```
klemens-morgenstern, line 132: I personally would do this: 
```suggestion
    int mib[4] = {CTL_KERN, KERN_PROC, KERN_PROC_PATHNAME, pid};
```

include/boost/process/v2/impl/exe.ipp
```diff
@@ -0,0 +1,324 @@
...
+        std::string strbuff;
+        strbuff.resize(len, '\0');
+        char *exe = strbuff.data();
+        if (sysctl(mib, 4, exe, &len, nullptr, 0) == 0) 
+        {
+            char buffer[PATH_MAX];
+            if (realpath(exe, buffer)) 
+            {
+                path = buffer;
+            }
```
klemens-morgenstern, line 149: ^ error handling?

include/boost/process/v2/impl/exe.ipp
```diff
@@ -0,0 +1,324 @@
...
+filesystem::path exe_path(boost::process::v2::pid_type pid, std::error_code & ec)
+{
+    std::string path;
+    int mib[4]; 
+    std::size_t len;
+    mib[0] = CTL_KERN;
+    mib[1] = KERN_PROC_ARGS;
+    mib[2] = pid;
+    mib[3] = KERN_PROC_PATHNAME;
+    if (sysctl(mib, 4, nullptr, &len, nullptr, 0) == 0)
```
klemens-morgenstern, line 168: this looks like it's the same as above? Why not just put the #if around the `mib` ?

include/boost/process/v2/impl/exe.ipp
```diff
@@ -0,0 +1,324 @@
...
+                        st.st_ino == static_cast<ino_t>(kif[i].va_fileid))) 
+                    {
+                        *out = executable;
+                        success = true;
+                        break;
+                    }
+                }
+            }
+            kvm_close(kd);
+        }
```
klemens-morgenstern, line 226: error handling !

include/boost/process/v2/impl/exe.ipp
```diff
@@ -0,0 +1,324 @@
...
+                is_exe = is_executable(pid, argv0.c_str(), &path, ec);
+            } 
+            else if (buffer[0].find('/') == std::string::npos) 
+            {
+                std::string penv = envvar_value_from_proc_id(pid, "PATH");
+                if (!penv.empty()) {
+                    std::vector<std::string> env;
+                    std::string tmp;
+                    std::stringstream sstr(penv); 
+                    while (std::getline(sstr, tmp, ':'))
```
klemens-morgenstern, line 250: That won't do. I got a whole set of facilities in the `environment` header to do all this stuff without the overhead of `string`  and `stringstream`. Please look there and minimize the amount of allocations.

include/boost/process/v2/impl/exe.ipp
```diff
@@ -0,0 +1,324 @@
...
+    if (path.empty())
+        ec = detail::get_last_error();
+    return path;
+}
+
+#elif defined(__sun)
+
+filesystem::path exe_path(boost::process::v2::pid_type pid, std::error_code & ec)
+{
+    std::string path;
```
klemens-morgenstern, line 295: why the intermediate `path` string btw.? WHy not just return `exe`?

include/boost/process/v2/impl/pid.ipp
```diff
@@ -22,6 +65,603 @@ pid_type current_pid() {return ::GetCurrentProcessId();}
...
+        ec = detail::get_last_error();
+        return vec;
+    } 
+    else
+    {
+        std::size_t size = fread(buffer, sizeof(char), sizeof(buffer), stat);
+        if (size > 0) 
+        {
+            char *token = nullptr;
+            if ((token = strtok(buffer, " "))) {
```
klemens-morgenstern, line 237: this if thingy is hard to read, can you organize that differently?

include/boost/process/v2/impl/pid.ipp
```diff
@@ -22,6 +65,603 @@ pid_type current_pid() {return ::GetCurrentProcessId();}
...
+                if ((token = strtok(nullptr, " "))) {
+                    if ((token = strtok(nullptr, " "))) {
+                        if ((token = strtok(nullptr, " "))) {
+                            pid_type ppid = (pid_type)strtoul(token, nullptr, 10);
```
klemens-morgenstern, line 241: `static_cast`

include/boost/process/v2/impl/pid.ipp
```diff
@@ -22,6 +65,603 @@ pid_type current_pid() {return ::GetCurrentProcessId();}
...
+                            vec.push_back(ppid);
+                        }
+                    }
+                }
+            }
+            if (!token)
+            {
+                ec = detail::get_last_error();
+                return vec;
```
klemens-morgenstern, line 250: that would leave `stat` open, wouldn't it?

include/boost/process/v2/impl/pid.ipp
```diff
@@ -22,6 +65,603 @@ pid_type current_pid() {return ::GetCurrentProcessId();}
...
+
+std::vector<pid_type> all_pids(std::error_code & ec)
+{
+    std::vector<pid_type> vec;
+    int cntp = 0;
+    kinfo_proc *proc_info = kinfo_getallproc(&cntp);
+    if (proc_info) 
+    {
+        for (int i = 0; i < cntp; i++)
+            vec.push_back(proc_info[i].ki_pid);
```
klemens-morgenstern, line 282: technically a data leak, if `push_back` throws, `free` won't get called.

Also: call `vec.reserve()` in all the examples please, so we only alloc once.

klemens-morgenstern: A smart_ptr might also do - `unique_ptr` with a custom deleter will do the trick.

include/boost/process/v2/impl/pid.ipp
```diff
@@ -22,6 +65,603 @@ pid_type current_pid() {return ::GetCurrentProcessId();}
...
+        {
+            ec = detail::get_last_error();
+            break;
+        }
+    }
+    kvm_close(kd);
+    return vec;
+}
+
+std::vector<pid_type> parent_pid(pid_type pid, std::error_code & ec) {
```
klemens-morgenstern: why is this returning a vector and not just one `pid_type` ?

test/v2/pid.cpp
```diff
@@ -7,9 +7,38 @@
...
+        {
+            std::vector<bp2::pid_type> tmp1;
+            std::vector<bp2::pid_type> tmp2 = bp2::child_pids(children[i]);
+            tmp1.insert(std::end(tmp1), std::begin(tmp2), std::end(tmp2));
+            grand_children = tmp1;
+        }
+        return (!children.empty() || !grand_children.empty());
+    };
+    #if defined(_WIN32)
+    // pid of zero is guaranteed to exist on Windows
```
klemens-morgenstern, line 38: this should be unified!

klemens-morgenstern: Could be that you have a `root_pid` constant that is either 1 or 0 depending on the os.

test/v2/pid.cpp
```diff
@@ -7,9 +7,38 @@
...
+    BOOST_CHECK_NE(grand_child_pids(0, children, grand_children), false);
+    #else
+    // pid of one is guaranteed to exist on Unix-likes
+    BOOST_CHECK_NE(grand_child_pids(1, children, grand_children), false);
+    #endif
+}
```
klemens-morgenstern, line 44: Is there a test to check for a child process we launched?

klemens-morgenstern (changes requested):

include/boost/process/v2/impl/exe.ipp
time-killer-games: if the path is empty, and path will only be set if realpath is not null, then it will throw an error as can be seen below.

I'm confused by your former comment regarding my realpath error handling being incorrect, the path will not be set unless realpath returned something besides nullptr, however i did what you wanted for linux anyway.

include/boost/process/v2/impl/exe.ipp
time-killer-games: I see now, the problem you had earlier with my realpath error handling was related to the fact i am copying the strings, which is more memory being allocated when it could be avoided. I'll fix that, nice catch!

test/v2/pid.cpp
time-killer-games: Yeah, that's a good point. I'll create a root_pid() function then.

test/v2/pid.cpp
time-killer-games: That would probably be better than what I have now, actually, because this test fails inconsistently where it stands.

include/boost/process/v2/impl/exe.ipp
time-killer-games: lol will do

include/boost/process/v2/impl/exe.ipp
time-killer-games: as i mentioned in another comment, checking against whether the returned string is empty is a waste of allocation, and a lot of memory being wasted just to half-ass the error handling, so yeah ill fix it. Thanks!

include/boost/process/v2/impl/pid.ipp
time-killer-games: correcto, nice catch

include/boost/process/v2/impl/pid.ipp
time-killer-games: in the case of error, it shouldn't return a pid at all. If you have a suggestion on an alternate way of handling that, I'm all ears.

include/boost/process/v2/detail/impl/exe.ipp
time-killer-games: If linux has anything like this it is bound to be provided by the procfs. A quick google search didn't give me any results i was hoping for, but i can look more after addressing more pressing things you critiqued :)

include/boost/process/v2/detail/impl/exe.ipp
time-killer-games: Not to mention that basic_process? What do you mean by that? I added a few lines giving explanation comments.

include/boost/process/v2/detail/impl/process_handle_windows.ipp
time-killer-games: I'm not sure if GetLastError returns with these undocumented api's. Isn't that what the NTSTATUS is for?

include/boost/process/v2/impl/pid.ipp
klemens-morgenstern: I think `-1` is fair - the value is nonsense if `ec` got set.

include/boost/process/v2/impl/pid.ipp
klemens-morgenstern: no, something like `std::pmr::polymorphic_allocator`. But I am unsure that's worth it here.

include/boost/process/v2/detail/impl/exe.ipp
klemens-morgenstern: I probably should've finished the sentence - the basic_process you get as result of launching already has the debug privilege I think.

include/boost/process/v2/detail/impl/process_handle_windows.ipp
klemens-morgenstern: I think so - and we'd need an error_category for NTSTATUS then, which may end up horrific.

include/boost/process/v2/impl/pid.ipp
time-killer-games: -1 is not possible on windows because it is unsigned long. although we could do static_cast<pid_type>(-1) which i guess is basically what you meant

include/boost/process/v2/detail/impl/exe.ipp
time-killer-games: well that's only if you launched it yourself it sounds like you meant, however this can access any process even ones you didn't run yourself. It allows you to read information from processes run by other users as well as admin run processes.

include/boost/process/v2/detail/impl/process_handle_windows.ipp
time-killer-games: yeah - i wouldn't overthink it, i would prefer being less paranoid for now and be satisfied with something less than perfect and fix it as needed if someone reports breakage or flaws, i as the only real way to know how to perfect it would be through hands on test cases imo

include/boost/process/v2/impl/pid.ipp
time-killer-games: Are you saying we should use `std::unique_ptr<pid_type>` instead of `std::vector<pid_type>`, correct?

include/boost/process/v2/impl/cwd.ipp
```diff
@@ -0,0 +1,248 @@
...
+    wchar_t cwd[MAX_PATH];
+    HANDLE proc = boost::process::v2::detail::ext::open_process_with_debug_privilege(pid, ec);
+    if (proc == nullptr) return path;
+    if (is_x86_process(GetCurrentProcess(), ec) != is_x86_process(proc, ec)) {
+        CloseHandle(proc); 
+        return "";
+    } else {
+      goto err;
+    }
+    boost::process::v2::detail::ext::cwd_cmd_env_from_proc(proc, &buffer, MEMCWD, ec);
```
klemens-morgenstern, line 76: I think this should return a buffer in a `vector`.

klemens-morgenstern:
Found it: https://learn.microsoft.com/en-us/windows/win32/api/winternl/nf-winternl-rtlntstatustodoserror

^ this needs to be used when we get an NTSTATUS as return value.

time-killer-games:
> Found it: https://learn.microsoft.com/en-us/windows/win32/api/winternl/nf-winternl-rtlntstatustodoserror
> 
> ^ this needs to be used when we get an NTSTATUS as return value.

Nice find!!!! :D

include/boost/process/v2/detail/impl/process_handle_windows.ipp
time-killer-games: I guess this can be marked as resolved as soon as I use your suggestion to convert NTSTATUS to a 'normal' win32 error code. I'll let you know once I've done this.

include/boost/process/v2/impl/pid.ipp
time-killer-games: let me know once you've made up your mind on this

klemens-morgenstern: yeah it's nonsense. we're iterating through the process table, a few allocs won't matter.

PR #291, opened 2022-12-29
fix error message

garymm:


klemens-morgenstern:
My obvious copy-pasta was obvious. thanks for the fix.

PR #307, opened 2023-04-23
Fix MSVC compile errors

nikola-sh:


include/boost/process/v2/environment.hpp
```diff
@@ -501,8 +501,8 @@ struct key
 
     template< class Source >
     key( const Source& source,
-        decltype(source.data()) = nullptr,
```
nikola-sh, line 504: error C2065: 'source': undeclared identifier

Tectu:
I encountered the same build issue using MinGW (via MSYS2).
I can confirm that this patch fixes the compilation errors.

klemens-morgenstern:
Thanks!

Issue #322, opened 2023-06-28
Setting the capacity of the underline pipe. Linux only


SilverPlate3:
I work for a big security vendor, and recently we tried to shift from libc interface (fork, exec, waitpid...) to Boost::Process.
However, we are missing some crucial API's related to the underline pipe.
One of the functionality we need is: Setting the pipe capacity. 
If the pipe fills up, 2 things may happen (Depends if the pipe is nonblocking or blocking):
1) **The producer will be blocked**. From [pipe man page](https://man7.org/linux/man-pages/man7/pipe.7.html): "If a process attempts to write to a full pipe, then blocks until sufficient data has been read from the pipe to allow the write to complete"
2) **The new data will be thrown**. From [pipe man page](https://man7.org/linux/man-pages/man7/pipe.7.html): "If the pipe is full, then write(2) fails, with errno set to EAGAIN."

Both cases cause us to loose live data from the producer. 
increasing the pipe capacity doesn't solve this issue 100%. But the possibility of this happening, diminishing by a lot.
This is crucial for us. 

**Running the following code for 5 seconds:**
```
int main()
{
    bp::ipstream stdOut;
    bp::child childProcess = bp::child("sysdig", bp::std_out > stdOut);
    auto fdStdOut = stdOut.pipe().native_source();
    
    while (true)
    {
        int bytesInPipe;
        ioctl(fdStdOut, FIONREAD, &bytesInPipe); // Find number of available bytes
        std::string output(bytesInPipe + 1, 0);
        std::cout << "\nPipe size: " << fcntl(fdStdOut, F_GETPIPE_SZ) << std::endl;
        std::cout << "Bytes available: " << bytesInPipe << std::endl;
        std::cout << "Bytes read: " << read(fdStdOut, output.data(), bytesInPipe) << std::endl;
        std::this_thread::sleep_for(std::chrono::milliseconds(50)); // Some work
    }
}
```

**Here is some of the output**:
```
Pipe size: 65536
Bytes available: 17685
Bytes read: 17685

Pipe size: 65536
Bytes available: 5577
Bytes read: 5577

Pipe size: 65536
Bytes available: 65536
Bytes read: 65536

Pipe size: 65536
Bytes available: 65421
Bytes read: 65421

Pipe size: 65536
Bytes available: 4564
Bytes read: 4564
```
The capacity is easily reached. 

After changing the pipe capacity to the value specified at: /proc/sys/fs/pipe-max-size, I was able to 20X the pipe capacity. 
[...]

PR #323, opened 2023-06-28
Set underline pipe capacity. Linux only. Issue - https://github.com/boostorg/process/issues/322

SilverPlate3:
Possible Solution to https://github.com/boostorg/process/issues/322.
Including unit tests. 

Please let me know if:
1) You prefer the API to be part of the bp::opstream/bp::ipstream. So users need to go into ip.pipe.set_pipe_capacity().
2) Is it ok I included <string> and <fstream>?   I saw you use them only in the unit tests. However because of the #if, I think it won't cause compilation errors. 
3) Would you like me to add it to the doc's?
4) When could it be pushed in?

### DIFF
--- a/include/boost/process/detail/posix/basic_pipe.hpp
+++ b/include/boost/process/detail/posix/basic_pipe.hpp
@@ -19,6 +19,12 @@
 #include <fcntl.h>
 #include <memory>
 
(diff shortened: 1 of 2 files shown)


SilverPlate3:
Please let me know how can I improve my PR, to better suite the code base.

PR #324, opened 2023-07-01
Set pipe size issue322

SilverPlate3:
Possible Solution to https://github.com/boostorg/process/issues/322.
Including unit tests.

Please let me know if:

1. You prefer the API to be part of the bp::opstream/bp::ipstream. So users won't need to go into ip.pipe.set_pipe_capacity().
2. Would you like me to add it to the doc's?
3. When could it be pushed in?


klemens-morgenstern:
I don't think this should be part of the API. It's linux only , so you can just use fcntl manually. Its trivial, really.

Issue #285, opened 2022-12-15
Zombie process created


reddwarf69:
In https://github.com/boostorg/process/commit/bbabea30dd0940ab604e332a50b38013138dd038 the handling of execve() errors was supposedly fixed.
But if I run
```
#include <boost/process.hpp>
#include <unistd.h>

int main() {
	try {
	 	boost::process::child("doesnt-exist");
	}catch(...){}

	sleep(100);
}
```
via strace, I can't see any wait() call
 ```$ strace -f -e process ./test
execve("./test", ["./test"], 0x7ffe241129e8 /* 11 vars */) = 0
clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLD, child_tidptr=0x7ff79b6ca6d0) = 84374
strace: Process 84374 attached
[pid 84374] execve("doesnt-exist", ["doesnt-exist"], 0x7ffd4520dd58 /* 11 vars */) = -1 ENOENT (No such file or directory)
[pid 84374] exit_group(1)               = ?
[pid 84374] +++ exited with 1 +++
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=84374, si_uid=1000, si_status=1, si_utime=0, si_stime=0} ---
```
And running top (and pressing `V` to get a process tree) while the program is in the `sleep(100);` confirms there is a zombie process.

I'm using Boost 1.81.0.

By the way, in https://github.com/boostorg/process/blob/a9f083f45c6da4273885d5869c1a55c1c7d19c7a/test/error.cpp#L113 and https://github.com/boostorg/process/blob/a9f083f45c6da4273885d5869c1a55c1c7d19c7a/test/error.cpp#L122, "doesnt-exit" probably is supposed to be "doesnt-exist"?

SilverPlate3:
@klemens-morgenstern 
Issue is fixed and tested in pull request https://github.com/boostorg/process/pull/325
I think issue can be closed.

PR #325, opened 2023-07-02
Reap zombie after failed execv issue285

SilverPlate3:
**Fixing issue** - https://github.com/boostorg/process/issues/285

**What was the problem**: After a failed execve, the caller was suppose to call waitpid and reap the zombie that the fork created.
However an exception was thrown before waitpid was called. 
This caused the child fork to become a zombie. 

Running the following code:
image
Would result:
image

**What I did**:
1) Extracting set_error() out of _read_error(). set_error() throws the exception, and is a big side effect for a function with the name of _read_error(). 
2) Calling set_error() Only after the waitpid. 
3) Attempting to call waitpid() twice, with a sleep between (if needed). The first waitpid may suffer a race condition, as the child fork didn't exit yet. 
4) Unit test.

### DIFF
--- a/include/boost/process/detail/posix/executor.hpp
+++ b/include/boost/process/detail/posix/executor.hpp
@@ -246,10 +246,14 @@ class executor
             //actually, this should block until it's read.
             auto err = errno;
             if ((err != EAGAIN ) && (err != EINTR))
-                set_error(std::error_code(err, std::system_category()), "Error read pipe");
+            {
+                _msg = "Error read pipe";
+                _ec = std::error_code(err, std::system_category());
+                return;
+            }
         }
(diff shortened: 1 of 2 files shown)


SilverPlate3:
@klemens-morgenstern 

klemens-morgenstern:
I don't think this fix does it the right way tbh.

SilverPlate3:
@klemens-morgenstern 
Can you please provide me with more information for why "this fix does it the right way"? Or a code review?
I would love to do it the right way! So please let me know what is right way in your eyes

klemens-morgenstern:
because a random sleep in a loop is the wrong choice.

PR #327, opened 2023-07-02
Align nonexistent binary name with other tests

SilverPlate3:
All the other tests that try to run a nonexistent binary use the name "doesnt-exist", which is a much more logical name to this purpose then "doesnt-exit".
For example:
https://github.com/boostorg/process/blob/bfb1ebb5bd24837b6292421dba3790f5701883ba/test/posix_specific.cpp#L96


klemens-morgenstern:
Lol, yeah you're right. To be fair, if it doesn't exist it also doesn't exit.

SilverPlate3:
@klemens-morgenstern  
I see 10 checks have failed in the CI/CD. Is it normal?
Surely its unrelated to these cosmetics changes.

SilverPlate3:
@klemens-morgenstern 

klemens-morgenstern:
I don't think how this fixes anything, it's a name change of a nonexisting binary.

SilverPlate3:
@klemens-morgenstern 
Indeed it doesn't fix any behavior, but it aligns naming with the rest of the code + gives a more "correct" name.
Clean code requires naming guide lines

Issue #256, opened 2022-05-31
Empty argument "disappears" on Windows


Capital-Z:
Passing empty argument to a process just adds an extra space to its command line on Windows which is normally ignored. Not sure how it would work on Posix, but makes it impossible to pass empty value under Windows.
Possible solution would be in ...process\detail\windows\basic_cmd.hpp build_args(..) which quotes arguments if necessary, to also quote empty ones. E.g. something like:
`if(!arg.empty())
{...}
else
arg = "\"\"";`

SilverPlate3:
@Capital-Z
Can you please

post the code you tried to run.
Elaborate on the desired outcome.
I wasn't able to fully understand the issue.
Thanks!

Capital-Z:
If I remember correctly the original problem was with a custom executable and I solved it by passing data to it in other means. But the issue is simple: how to pass empty argument?
As example ImageMagic has option -format and if you want to suppress it you pass empty string for format, i.e. from terminal: `magic.exe -format ""`
So in boost process terms it would be (not verified code just typing here):
`boost::process::child("magic.exe", boost::process::args({"-format", ""}));`
According to my original comment this would pass just "-format" to magic.exe, which is not valid input for it.
I am not sure if my original digging in boost/process code is still relevant.

SilverPlate3:
@klemens-morgenstern 
Fixed the issue in pull request https://github.com/boostorg/process/pull/328/files
I think issue can be closed.

fkonvick:
@klemens-morgenstern @SilverPlate3 This broke existing code for us.  We used to:
`bp::child(bp::shell, bp::args = {shell_script}, ...`
which resulted in `C:\Windows\system32\cmd.exe /c shell_script` out of `build_args` or, in case there was a space or `"` in the shell script, a quoted/escaped version: `C:\Windows\system32\cmd.exe /c "shell_script"`.  See PR #73 for fixes that were earlier introduced to fix escaping.

With this change in PR @256, we're getting
`C:\Windows\system32\cmd.exe /c "" shell_script` (note the extra empty pair of double quotes) and the script refuses to run.

I see that the co-culprit is `exe_args_shell`:  this code
```
std::vector<string_type> args_ = {c_arg(Char()), std::move(exe)};
args_.insert(args_.end(), std::make_move_iterator(args.begin()), std::make_move_iterator(args.end()));
```
[...]

PR #328, opened 2023-07-05
Pass empty argument issue256

SilverPlate3:
### Fixing issue: https://github.com/boostorg/process/issues/256
At the moment, we can't pass an empty argument "" to the child. This is an issue for programs that may take an empty string as an argument. 
**This limitations only occurs on windows!**

For example a string replacement program. You might want to replace "XYZ" with "" (thus deleting the unwanted string).
In the above issue, the user needed it for the ImageMagic app.

I created a simple program that takes arguments via CMD and prints them
```
int main(int argc, char* argv[]) {
    std::cout << "Number of arguments: " << argc - 1 << std::endl;
    for (int i = 1; i < argc; ++i) {
        std::cout << "Argument " << i << ": " << argv[i] << std::endl;
    }
    std::cout << "\n" << std::endl;
}
```
Running the program with an empty argument via cmd:
image

Current boost::process implementation:
When I try to pass an empty string as an argument, it gets "thrown" and isn't passed properly to the child program. 
image
image

After changing the code so it will pass empty arguments properly:
image

Good thing is that looks like we can't pass an un-intended empty argument (we must specify ""). Or at least I wasn't able to reproduce such edge case.


SilverPlate3:
@klemens-morgenstern

klemens-morgenstern:
Master branch is locked for thebcurrent releasr atm.

PR #330, opened 2023-07-11
Use boost::throw_exception rather than direct throw

edtanous:
Using boost::throw_exception allows for modifications to these exceptions on a per-application basis, including overriding with custom implementations.

This also has the benefit of allowing compilation with -fno-exceptions set, which should make this code more portable.

### DIFF
--- a/include/boost/process/detail/config.hpp
+++ b/include/boost/process/detail/config.hpp
@@ -21,7 +21,9 @@
 #include <system_error>
 #include <boost/system/api_config.hpp>
 
+#include <boost/throw_exception.hpp>
 #include <boost/process/exception.hpp>
+#include <boost/assert/source_location.hpp>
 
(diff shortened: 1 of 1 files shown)


edtanous:
> Can you add the boost::source_location too? I'll merge then.

Do you mean as a #include, or would you like me to change each of these methods to use the overload of boost::throw_exception that accepts a source_location as an argument?

edtanous:
> Can you add the boost::source_location too? I'll merge then.

I've made my best guess as to what I think you're asking for here.  Please take a look.

klemens-morgenstern:
Looks good, although I'd also add it to the throw_last_error functions.

edtanous:
> Looks good, although I'd also add it to the throw_last_error functions.

Done.

edtanous:
Thank you.  I appreciate you.

PR #333, opened 2023-08-10
Fix compilation with -Wall and -Werror

Kalabaza:
While compiling with gcc 11.4 using -Wall and -Werror, this warning stops the compilation:

build/_deps/boost-src/libs/process/include/boost/process/detail/posix/async_out.hpp:74:38: error: unused parameter ‘exec’ [-Werror=unused-parameter]
   74 |     inline void on_success(Executor &exec)
      |                            ~~~~~~~~~~^~~~
cc1plus: all warnings being treated as errors


klemens-morgenstern:
If everything passes I'll merge after the boost release.

Issue #217, opened 2021-08-19
boost::process:child::wait() on Linux ignores a zombie process 


theroyn:
Reproducible in the following example:
```
#include <boost/process.hpp>

int main()
{
        namespace bp = boost::process;

        std::string appCommand = "sleep 50";
        bp::child childProcess(appCommand);

        int pid = childProcess.id();

        printf("childProcess.running()=%d\n", childProcess.running());

        childProcess.terminate();
        childProcess.wait();

        printf("childProcess.running()=%d\n", childProcess.running());

        // kill 0 returns 0 if process is running
        bool isRunning = (::kill(pid, 0) == 0);

        printf("isRunning=%d\n", isRunning);
}
```
output on my machine is:
```
childProcess.running()=1
childProcess.running()=0
(kill 0 23764): isRunning=1
```
The program will print ```isRunning=1``` at its end, meaning that the process is still running. Pausing at that point with debugger and checking ```ps -ax``` will also show that the process is a zombie.

After some investigation, faulty code seem to be in ```include/boost/process/detail/child_decl.hpp```:
```
    bool _exited()
    {
        return _terminated || !::boost::process::detail::api::is_running(_exit_status->load());
    };

    void terminate(std::error_code & ec) noexcept
    {
        if (valid() && running(ec) && !ec)
            boost::process::detail::api::terminate(_child_handle, ec);

        if (!ec)
            _terminated = true;
    }

    void wait(std::error_code & ec) noexcept
    {
        if (!_exited() && valid())
        {
            int exit_code = 0;
            boost::process::detail::api::wait(_child_handle, exit_code, ec);
            if (!ec)
                _exit_status->store(exit_code);
        }
    }
```
```terminate()``` marks ``` _terminated = true```, and ```wait()``` never calls its main logic because ```_exited()``` evaluates true.

A possible solution would be to perform a blocking wait at the end of ```boost::process::detail::api::terminate()```, because ATM it waits with ```WNOHANG```.

SilverPlate3:
@klemens-morgenstern 
Do you want me to create a pull request and fix this issue?

klemens-morgenstern:
I think I fixed this, maybe add a test for it?

jmccabe:
Should this be closed and, if so, has it been fixed at some point or just replaced? child_decl.hpp in the now v1 directory on the develop branch hasn't had any updates to do anything about this.

PR #337, opened 2023-08-21
Reap zombie after failed execve. Issue 285 

SilverPlate3:
After @klemens-morgenstern logical comments on the latest [pull request](https://github.com/boostorg/process/pull/325).
I created this pull request and closed the previous one.

In order to fix the [issue #217](https://github.com/boostorg/process/issues/285) this pull request does:
1 - Extracts ` _set_error()` out of `_read_error()`.  The function  `_set_error()` throws an exception which causes the code to never get to the `waitpid()` that cleans a zombie process after it failed an execve. 
Now `_read_error()` only reads the error as its name implies, and `_set_error()` will be called only after the `waitpid()`

2 - After the first step, I faced a race condition between the `::waitpid(this->pid, nullptr, WNOHANG);` and the actual `_exit(EXIT_FAILURE);` of the process which failed the execve. The `::waitpid(this->pid, nullptr, WNOHANG);` was called some times before the `_exit(EXIT_FAILURE);`.
In order to fix this, I check what was the failure reason that I get from `_read_error(p.p[0]);` and set the `waitpid()` parameter according to it. 

3 - Added a unit test to check that the fix works.

### DIFF
--- a/include/boost/process/detail/posix/executor.hpp
+++ b/include/boost/process/detail/posix/executor.hpp
@@ -246,10 +246,14 @@ class executor
             //actually, this should block until it's read.
             auto err = errno;
             if ((err != EAGAIN ) && (err != EINTR))
-                set_error(std::error_code(err, std::system_category()), "Error read pipe");
+            {
+                _msg = "Error read pipe";
+                _ec = std::error_code(err, std::system_category());
+                return;
+            }
         }
         if (count == 0)
-            return  ;
+            return;
 
         std::error_code ec(data[0], std::system_category());
         std::string msg(data[1], ' ');
@@ -262,9 +266,14 @@ class executor
                 return;
                 //EAGAIN not yet forked, EINTR interrupted, i.e. try again
             else if ((err != EAGAIN ) && (err != EINTR))
-                set_error(std::error_code(err, std::system_category()), "Error read pipe");
+            {
+                _msg = "Error read pipe";
+                _ec = std::error_code(err, std::system_category());
+                return;
+            }
         }
-        set_error(ec, std::move(msg));
+        _msg = std::move(msg);
+        _ec = ec;
     }
 
     std::string prepare_cmd_style_fn; //buffer
@@ -434,7 +443,6 @@ child executor<Sequence>::invoke(boost::mpl::false_, boost::mpl::false_)
             ::close(p.p[1]);
 
             _exit(EXIT_FAILURE);
-            return child();
         }
 
         ::close(p.p[1]);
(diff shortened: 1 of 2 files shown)


include/boost/process/detail/posix/executor.hpp
```diff
@@ -445,8 +453,10 @@ child executor<Sequence>::invoke(boost::mpl::false_, boost::mpl::false_)
     if (_ec)
     {
         //if an error occurred we need to reap the child process
-        ::waitpid(this->pid, nullptr, WNOHANG);
+        int block_mode = _msg == "execve failed" ? 0 : WNOHANG;
```
klemens-morgenstern, line 456: Why is this conditional?

include/boost/process/detail/posix/executor.hpp
```diff
@@ -445,8 +453,10 @@ child executor<Sequence>::invoke(boost::mpl::false_, boost::mpl::false_)
...
+        ::waitpid(this->pid, nullptr, block_mode);
         boost::fusion::for_each(seq, call_on_error(*this, _ec));
+        set_error(_ec, _msg);
```
klemens-morgenstern, line 459: what is this for? we already reported the error.

include/boost/process/detail/posix/executor.hpp
SilverPlate3: Great question. 
**First we must know**:
Doesn't Block - `::waitpid(this->pid, nullptr, WNOHANG); `
Blocks until it has a child zombie to reap - `::waitpid(this->pid, nullptr, 0); `

We don't always want to block. For example, if we look in `_read_error()`, there are several things that can assign a value to `_ec`. One of them is a failed `::read()`. Failed `::read()` doesn't ensure us that we have a zombie child to reap. We want to use  `::waitpid(this->pid, nullptr, 0); ` only when we are sure we have a zombie to reap so we won't be blocked for a long time.
Only if `_msg == "execve failed"` we know the child has exited or will exit in its next few lines.

include/boost/process/detail/posix/executor.hpp
SilverPlate3: Indeed. 

However, on top of setting `_ec` and `_msg`, `set_error()` indirectly calls `throw process_error(ec, msg); `
Because I have extracted `set_error()` out of `_read_error() ` and because `throw process_error` is only called in error related functions in executor.hpp I thought it will be most appropriate. 

Would you prefer me changing this `set_error(_ec, _msg);` to `throw process_error(_ec, _msg); ` ?

SilverPlate3:
@klemens-morgenstern Hey.
When can I expect this to be merged? Or a respond on the pull request comments?
Thanks you very much!

include/boost/process/detail/posix/executor.hpp
klemens-morgenstern: I don't think that's correct. The closed pipe would indicate that the child process exited already, so WNOHANG should be correct.

include/boost/process/detail/posix/executor.hpp
klemens-morgenstern: Neither is correct I fear.

include/boost/process/detail/posix/executor.hpp
klemens-morgenstern: And comparing error messages is just bad practice. Why can't we just wait and ignore the error if the child doesn't exists?

klemens-morgenstern:
Can you check if 13c4b0aa144714007bfe26ab1a5db2e0f1519011 solves the issue?

PR #354, opened 2024-02-06
Add more error handling to bp2::ext cwd.ipp.

time-killer-games:


time-killer-games:
something tells me i didn't do this correctly at all :P @klemens-morgenstern

time-killer-games:
@klemens-morgenstern also wanted to mention, `procstat_freefiles()`, `procstat_freeprocs()`, and `procstat_close()` all return void and seem to have no defined error reporting mentioned in the FreeBSD documentation, all the more reason for us to switch to using `libkvm` like I mentioned to you recently in a private message on the cpp lang slack.

For reference: https://man.freebsd.org/cgi/man.cgi?query=libprocstat&sektion=3&format=html

include/boost/process/v2/ext/impl/cwd.ipp
```diff
@@ -185,7 +187,8 @@ filesystem::path cwd(boost::process::v2::pid_type pid, boost::system::error_code
         }
         else
             BOOST_PROCESS_V2_ASSIGN_LAST_ERROR(ec)
-        pclose(fp);
+        if (pclose(fp) == -1)
```
klemens-morgenstern, line 190: That looks to be correct.

include/boost/process/v2/ext/impl/cwd.ipp
```diff
@@ -139,19 +138,23 @@ filesystem::path cwd(boost::process::v2::pid_type pid, boost::system::error_code
...
             else
                 BOOST_PROCESS_V2_ASSIGN_LAST_ERROR(ec)
             procstat_freeprocs(proc_stat, proc_info);
+            if (errno)
+                BOOST_PROCESS_V2_ASSIGN_LAST_ERROR(ec)
         }
         else
             BOOST_PROCESS_V2_ASSIGN_LAST_ERROR(ec)
         procstat_close(proc_stat);
+        if (errno)
```
klemens-morgenstern, line 156: Is this i a guess, or do you know this resets `errno` if it succeeds? Because it might just be a function that never fails.

time-killer-games: for other people reading this pull request who don't know what i told klemens on slack, these functions never fail and are just wrappers around the free() function, after inspecting FreeBSD's source tree for libprocstat's source code.

PR #376, opened 2024-05-30
Fixed close handles  with Alpine and MUSL

dchansen:
The MUSL C library does not provide the close_range in unistd.h and Alpine linux therefore cannot compile boost::process::v2, at close_handles.ipp fails to compile.

This PR adds a check to only use close_range if we are using GLIBC.

### DIFF
--- a/doc/v2/stdio.qbk
+++ b/doc/v2/stdio.qbk
@@ -12,7 +12,7 @@ automatically connected and the other side will get assigned to the child proces
 
 ```
     asio::io_context ctx;
-    asio::readable_pipe rp;
+    asio::readable_pipe rp{ctx};
 
     process proc(ctx, "/usr/bin/g++", {"--version"}, process_stdio{{ /* in to default */}, rp, { /* err to default */ }});
     std::string output;
--- a/include/boost/process/detail/posix/executor.hpp
+++ b/include/boost/process/detail/posix/executor.hpp
@@ -398,12 +398,16 @@ child executor<Sequence>::invoke(boost::mpl::false_, boost::mpl::false_)
             set_error(err, "fcntl(2) failed");//this might throw, so we need to be sure our pipe is safe.
             return child();
         }
+
+        _pipe_sink = p.p[1];
+
         _ec.clear();
         boost::fusion::for_each(seq, call_on_setup(*this));
 
         if (_ec)
         {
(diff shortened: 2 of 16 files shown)


klemens-morgenstern:
Looks good, will merge if all tests pass.

klemens-morgenstern:
Ok, I don't think this is correct, as it doesn't handle libc++. I'll need to add a MUSL test to CI.

t43rr7:
Hello, @dchansen, I've encountered the same issue, but with glibc. The `close_range` checks here would be a bit more complicated. I can include your changes in my [patch](https://github.com/boostorg/process/pull/378), so it could resolve 2 issues with single PR.

dchansen:
@t43rr7 That looks good to me. Could you make a PR for my branch? then I will merge it in.

t43rr7:
@dchansen I've opened PR to your branch. Looks like your branch's HEAD a bit behind origin develop, so there're a bit more changes that expected. Anyway, the most interesting diffs [here](https://github.com/dchansen/process/pull/1/files#diff-d2237ff6720cb6f4cd709d6300e6242b33f036ccb3b19bbb54768e99e1a976bbR14). I've also added raw system call support which should work even on MUSL library. 

Please, verify this changes on your setup, because I have no possibility to do it fast. Thanks!

t43rr7:
@klemens-morgenstern please, look at the new changes

klemens-morgenstern:
Can't we just put a `close_range` wrapper around the syscall into a detail namespace and use `using ::close_range` otherwise? That way we don't need to reimpl `close_all`.

t43rr7:
Do you mean something like "code below"?

```cpp

int close_range_wrapper(...) {
#ifdef HAS_CLOSE_RANGE
    return ::close_range(...);
#elif defined(HAS_CLOSE_RANGE_SYSCALL) 
    return ::syscall(SYS_close_range, ...);
#endif
}
```

t43rr7:
@klemens-morgenstern done in [this commit](https://github.com/boostorg/process/pull/378/commits/40350de7ebeb353178b9fa6b8bb9d0f0ef0b6efc)

Please, see [PR](https://github.com/boostorg/process/pull/378)

t43rr7:
@klemens-morgenstern are there any news?

klemens-morgenstern:
I think this is fixed in develop already..

PR #378, opened 2024-06-18
posix: check glibc version for ::close_range function

T3RR7:
On old systems with old glibc, but new kernel (the common case is build in CI on ancient OS with old glibc for extended compatibility) there may not be `::close_range()` function. In this case build will fail.

To avoid that we should properly check not only linux kernel version, but glibc version too.

There is one another option to solve this issue is use raw system call with only checking kernel version, but it seems to me dirty solution.

### DIFF
--- a/include/boost/process/v2/posix/detail/close_handles.ipp
+++ b/include/boost/process/v2/posix/detail/close_handles.ipp
@@ -45,12 +45,26 @@ int fdwalk(int (*func)(void *, int), void *cd);
 
 #include <linux/version.h>
 
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,11,0)
+#ifdef __GLIBC__
 
+#include <gnu/libc-version.h>
+
+#if __GLIBC__ >= 2 && __GLIBC_MINOR__ >= 34
+// on glibc we also must have ::close_range() wrapper around system call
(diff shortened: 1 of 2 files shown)


t43rr7:
Moreover, I've just noticed that kernel version also checked wrong. 
According to: 
- https://man.archlinux.org/man/close_range.2.en#HISTORY
- https://man7.org/linux/man-pages/man2/close_range.2.html

`close_range()` available from kernel 5.9, so right check might look like:

```cpp
#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,9,0)
```

t43rr7:
@dchansen I've seen you pushed the fix to you `devel` branch, but you haven't provided option to fully disable close_range usage, as I understood. 

In case we build on new kernel, BUT with old glibc (for extended compatibility) the whole program will break, because of ENOTSUP error (a.k.a Unsupported system call).

t43rr7:
@klemens-morgenstern  I've seen you pushed the fix to you devel branch, but you haven't provided option to fully disable close_range usage, as I understood.

In case we build on new kernel, BUT with old glibc (for extended compatibility) the whole program will break, because of ENOTSUP error (a.k.a Unsupported system call).

klemens-morgenstern:
@T3RR7 does 5a8102c work?

klemens-morgenstern:
Assuming 41a4a5a did the trick

PR #381, opened 2024-07-08
FIX: fix the incorrect parameter parsing

zhixingchen0629:
if i pass this params vector to launch process when the executor was nullptr:
```cpp
        std::string executor;
        std::vector<std::string> params;
        params.push_back("powershell");
        params.push_back("-noprofile");
        params.push_back("-executionpolicy");
        params.push_back("bypass");
        params.push_back("-file");
        params.push_back(m_command);
        auto launcher = boost::process::v2::default_process_launcher();
        m_child = std::move(launcher(m_ioContext, ec, executor.c_str(), params, boost::process::v2::process_stdio{nullptr, m_wPipe, m_wPipe}));
```
The actual splicing command_line at the bottom level is:
```
"" powershell -noprofile -executionpolicy bypass -file c:aaa
```

`"" `It is superfluous, which cause create process error.

parameter splicing should take into account the situation where executor is empty

### DIFF
--- a/include/boost/process/v2/posix/default_launcher.hpp
+++ b/include/boost/process/v2/posix/default_launcher.hpp
@@ -329,7 +329,7 @@ struct default_launcher
                     Args && args,
                     Inits && ... inits ) -> basic_process<typename ExecutionContext::executor_type>
     {
-        return (*this)(context.get_executor(), executable, std::forward<Args>(args), std::forward<Inits>(inits)...);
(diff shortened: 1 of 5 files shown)


klemens-morgenstern:
I am assuming that was fixed in develop already?

zhixingchen0629:
> assuming

sorry, my pr was wrong commit, please check my latest pr

PR #387, opened 2024-07-19
Added an Alpine Linux CI

anarthal:
As a regression test for #376 
Failing job: https://github.com/anarthal/process/actions/runs/10010298153/job/27671166551 (requires action beyond my knowledge).

PR #388, opened 2024-07-19
added CLOSE_RANGE_UNSHARE define to #387

klemens-morgenstern:


klemens-morgenstern:
Alpine passed, so it does the thing.

Issue #391, opened 2024-07-26
Process v2 does not build on OpenBSD


brad0:
Boost 1.86.0 beta1 on OpenBSD/amd64 -current.

```
common.mkdir bin.v2/libs/process/build/clang-linux-16/release/x86_64/link-static/pch-off/visibility-hidden/posix

        mkdir -p "bin.v2/libs/process/build/clang-linux-16/release/x86_64/link-static/pch-off/visibility-hidden/posix"

common.mkdir bin.v2/libs/process/build/clang-linux-16/release/x86_64/link-static/pch-off/visibility-hidden/windows

        mkdir -p "bin.v2/libs/process/build/clang-linux-16/release/x86_64/link-static/pch-off/visibility-hidden/windows"

clang-linux.compile.c++ bin.v2/libs/process/build/clang-linux-16/release/x86_64/link-static/pch-off/visibility-hidden/posix/close_handles.o

    "c++"   -fvisibility-inlines-hidden -Wall -fvisibility=hidden -Wno-inline -m64 -O2 -pipe   -O2 -pipe   -DBOOST_ALL_NO_LIB=1 -DBOOST_ASIO_NO_DEPRECATED -DBOOST_COBALT_USE_STD_PMR=1 -DBOOST_FILESYSTEM_STATIC_LINK=1 -DBOOST_PROCESS_SOURCE=1 -DNDEBUG   -I"."  -c -o "bin.v
2/libs/process/build/clang-linux-16/release/x86_64/link-static/pch-off/visibility-hidden/posix/close_handles.o" "libs/process/src/posix/close_handles.cpp"

libs/process/src/posix/close_handles.cpp:125:15: error: no member named 'close_range' in the global namespace
            ::close_range(0, whitelist.front() - 1, 0);
            ~~^
libs/process/src/posix/close_handles.cpp:135:19: error: no member named 'close_range' in the global namespace
                ::close_range(mine + 1, next - 1, 0);
                ~~^
2 errors generated.
clang-linux.compile.c++ bin.v2/libs/process/build/clang-linux-16/release/x86_64/link-static/pch-off/visibility-hidden/windows/default_launcher.o

[...]

PR #395, opened 2024-08-24
Fix building on OpenBSD

brad0:
OpenBSD does not have close_range() nor does NetBSD.

OpenBSD needs environ like the other *BSD's.

The build was erroring on kp_pid, it looks like p_pid is appropriate.

Fixes: https://github.com/boostorg/process/issues/391

### DIFF
--- a/include/boost/process/v2/posix/default_launcher.hpp
+++ b/include/boost/process/v2/posix/default_launcher.hpp
@@ -29,7 +29,7 @@
 #include <unistd.h>
 
 
-#if defined(__NetBSD__) || defined(__FreeBSD__) || defined(__APPLE__) || defined(__MACH__)
+#if defined(__NetBSD__) || defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__APPLE__) || defined(__MACH__)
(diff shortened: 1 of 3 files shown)


time-killer-games:
Hey, in case you were wanting to add other features to OpenBSD which are missing such as `bp2::ext::exe` you can base your code loosely on this file, just like I did with my past contributions to boost process v2:

https://github.com/time-killer-games/xproc/blob/main/process.cpp

Be sure to change the license from MIT to what boost process v2 has, I give you permission as the copyright holder of the relevent code. Process Hacker developers and devKathy are only responsible for small portions of code specific to the Windows platform, the rest I am the copyright holder for, which means I wrote all OpenBSD-related code, (and *BSD-related code as a whole). Make sure if you use enough of my code to be considered substantial that I am mentioned in the copyright of the related files (refer to me as "Samuel Venable").

I don't really have the time to add these features myself right now, but there is no obligation for you to do so either.

time-killer-games:
Does this actually build now? If not, let me know once you solve all build errors. @klemens-morgenstern would like me to test this pull request once you are done.

PR #396, opened 2024-08-30
Corrections to V2 docs

justend29:


klemens-morgenstern:
I think this is fixed now with the switch to asciidoc, can you check?

justend29:
@klemens-morgenstern The examples used in the docs/ of the develop branch have this MR's changes present. I'm not sure where to find the built documentation for the tip of develop, but since these examples are now compiled, I suspect they'll translate to docs correctly. Nice!

justend29:
I'll close this PR since the changes have been added as part of the general documentation updates.

PR #402, opened 2024-09-13
Documentation fixes

aluaces:
I think the last example is incorrect, since the text cites the `.interrupt` method but does not use it.


klemens-morgenstern:
Done

PR #410, opened 2024-10-03
[hotfix] Use correct definition for Boost ASIO when running on Windows

uilianries:
Hello!

The Boost.Asio project defines the macro `BOOST_ASIO_WINDOWS`, but not `ASIO_WINDOWS` as expected in this project. Reference:

https://github.com/boostorg/asio/blob/boost-1.86.0/include/boost/asio/detail/config.hpp#L713


klemens-morgenstern:
That's for standalone mode - what's going wrong?

PR #435, opened 2024-12-06
Suppress unused parameter warnings

Flamefire:


include/boost/process/v2/posix/default_launcher.hpp
```diff
@@ -49,10 +51,11 @@ struct base {};
 struct derived : base {};
 
 template<typename Launcher, typename Init>
-inline error_code invoke_on_setup(Launcher & launcher, const filesystem::path &executable,
+inline error_code invoke_on_setup(Launcher &, const filesystem::path &executable,
                                   const char * const * (&cmd_line),
-                                  Init && init, base && )
+                                  Init &&, base && )
 {
+    boost::ignore_unused(executable, cmd_line);
```
klemens-morgenstern, line 58: Why are you using ignore_unused for some and remove the names for others?

Flamefire: I found this more understandable this way. The name of the argument in `Launcher & launcher` isn't required but for `const filesystem::path &executable` it does help in understanding what would be here. Similar for `error_code& ec`

This allows easily verifying that the signature is correct.
The alternatives are to use either approach consistently which is more code, or commenting out the parameter name instead of removing it.

klemens-morgenstern:
Fixed in 662c0b10e3e46bca1c523c36b80511bd9c82d9ef

Flamefire:
Thanks!

PR #449, opened 2025-01-12
Fix POSIX `process_stdio` move semantics

### LINKED ISSUE #448: `v2::process_stdio` prematurely closes file descriptors when moved-from instance goes out of scope
On POSIX, the `v2::detail::process_io_binding` members of `v2::process_stdio` do not implement move semantics correctly: The underlying file descriptors are closed in the destructor if the `fd_needs_closing` flag is set, but neither the file descriptor, nor this flag, are reset when `process_io_binding` is moved.
https://github.com/boostorg/process/blob/b529769eb5d08d9f95b0130fd2f1a49acad330f4/include/boost/process/v2/stdio.hpp#L162-L166

I've encountered this as a problem when writing tests for some code that launches a process and reads some data from stdout. I wanted to avoid actually launching said process in tests, so I came up with a `FakeProcess` which I intend to be a drop-in replacement for `v2::process` in a generic wrapper. This is a small example demonstrating the issue:

```cpp
auto read_msg(boost::asio::io_context& context,
              boost::asio::readable_pipe read_pipe)
    -> boost::asio::awaitable<boost::system::error_code> {
    std::string s;
    boost::system::error_code ec;
    co_await boost::asio::async_read_until(
        read_pipe, boost::asio::dynamic_buffer(s), '\u0004',
        boost::asio::redirect_error(boost::asio::use_awaitable, ec));
    co_return ec;
}

auto write_msg(boost::asio::writable_pipe write_pipe)
    -> boost::asio::awaitable<void> {
    co_await boost::asio::async_write(write_pipe,
                                      boost::asio::buffer("Hello\u0004"));
}

struct FakeProcess {
    FakeProcess(boost::asio::io_context& context,
                boost::process::v2::process_stdio stdio)
        : _stdio{std::move(stdio)}
    {
        // Comment out the line below to "fix" the problem
        // stdio.out.fd_needs_closing = false;

        boost::asio::writable_pipe write_pipe{context, _stdio.out.fd};
        boost::asio::co_spawn(context, write_msg(std::move(write_pipe)),
                              boost::asio::detached);
    }

   private:
    boost::process::v2::process_stdio _stdio;
};

int main() {
    boost::asio::readable_pipe read_pipe{context};
    boost::process::v2::process_stdio stdio{.in = {}, .out = read_pipe, .err = {}};

    FakeProcess p(context, std::move(stdio));

    auto future = boost::asio::co_spawn(context,
                                        read_msg(context, std::move(read_pipe)),
                                        boost::asio::use_future);
    context.run();
    std::puts(future.get().message().c_str());
}
```
As is, this code prints "End of file" because the pipe is closed prematurely when `stdio` is moved into the member `_stdio`. [(Godbolt link)](https://godbolt.org/z/zTM75avEv)

Commenting out the line in the constructor of `FakeProcess` fixes the issue by toggling the flag in the moved-from object, thereby "defusing" its destructor. Clearly, that's just a dirty workaround (made possible by `process_io_binding`'s lack of encapsulation).

jgreitemann:
Fixes #448. Aside from adding the move operations, I've also added two test cases: one demonstrating implicit creation of pipes by `process_stdio` which was previously untested AFAICT, and one (regression) test specifically for the move semantics.

I'm curious if you'd also be open to making the members for `process_io_binding` private (on both platforms, actually)? `process_io_binding` has to maintain an invariant (whether `fd` needs to be closed) and leaving its members public allows external code to break that invariant. I'm not sure that your policy is towards breaking source compatibility in the `detail` namespace, so I have refrained from making that change.

### DIFF
--- a/include/boost/process/v2/stdio.hpp
+++ b/include/boost/process/v2/stdio.hpp
@@ -164,6 +164,31 @@ struct process_io_binding
   }
 
   process_io_binding() = default;
+  process_io_binding(const process_io_binding &) = delete;
+  process_io_binding & operator=(const process_io_binding &) = delete;
+
+  process_io_binding(process_io_binding && other) noexcept
+          : fd(other.fd), fd_needs_closing(other.fd), ec(other.ec)
+  {
+    other.fd = target;
+    other.fd_needs_closing = false;
+    other.ec = {};
+  }
+
+  process_io_binding & operator=(process_io_binding && other) noexcept
+  {
+    if (fd_needs_closing)
+      ::close(fd);
+
+    fd = other.fd;
+    fd_needs_closing = other.fd_needs_closing;
+    ec = other.ec;
+
+    other.fd = target;
+    other.fd_needs_closing = false;
+    other.ec = {};
+    return *this;
+  }
 
   template<typename Stream>
   process_io_binding(Stream && str, decltype(std::declval<Stream>().native_handle()) = -1)
--- a/test/v2/process.cpp
+++ b/test/v2/process.cpp
@@ -361,6 +361,43 @@ BOOST_AUTO_TEST_CASE(stdio_creates_complementary_pipes)
   BOOST_CHECK(proc.exit_code() == 0);
 }
 
+BOOST_AUTO_TEST_CASE(stdio_move_semantics)
+{
+  using boost::unit_test::framework::master_test_suite;
+  const auto pth =  master_test_suite().argv[1];
+
+  asio::io_context ctx;
+
+  asio::readable_pipe rp{ctx};
+  asio::writable_pipe wp{ctx};
+
+  auto make_stdio = [&]() -> bpv::process_stdio {
+      bpv::process_stdio stdio{};
+      stdio.in = wp;
+      stdio.out = rp;
+      stdio.err = nullptr;
+      // intentionally pessimizing move, preventing NRVO
+      return std::move(stdio);
+  };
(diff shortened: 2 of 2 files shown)


klemens-morgenstern:
Re private: unix users do interesting things with FDs that I cannot anticipate.  Modifying these member is discouraged by the detail namespace, so it is a bit lf a "I know what I am doing" if someone accesses this.

klemens-morgenstern (approved):
Looks good.

klemens-morgenstern:
FreeBSD is a broken CI setup.

klemens-morgenstern:
The windows test on GHA is funny:

```cpp
 libs/process/test/v2/ext.cpp(116): error: in "ext/test_cwd_exe": check tt == tmp has failed [C:\Users\runneradmin\AppData\Local\Temp != "C:\Users\RUNNER~1\AppData\Local\Temp"]
```

But unrelated to this PR, so not a blocker.

jgreitemann:
Thanks for reviewing this so timely.

I don't know about other issues in the FreeBSD pipeline, but it seems I made it worse because now there's a warning (as error) for the (intentional) pessimizing move.

klemens-morgenstern:
The freeBSD one was broken for non boost-process reasons (couldn't load DLL, some install mismatch).

There is a warning, but I honestly don't care for warnings in tests. If you want to fix that with a #pragma or some other way, I'm happy to merge.

PR #451, opened 2025-01-14
Remove optional dependency on filesystem when not meant to be

SimonMaracine:
Build previously failed when BOOST_PROCESS_USE_STD_FS was ON, because filesystem was linked to regardless.

I tried to use this library with the std version of filesystem and I noticed that Boost::filesystem was listed as a dependency regardless of the flag. I'm pretty sure this is just a little mistake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,7 +35,6 @@ target_link_libraries(boost_process
     Boost::asio
     Boost::config
     Boost::core
-    Boost::filesystem
     Boost::fusion
     Boost::iterator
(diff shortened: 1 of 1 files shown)


SimonMaracine:
I am very sorry. I just noticed that I have made the commit in the `develop` branch instead of a separate one. I haven't made a contribution in a long time. I think it's better to just close this pull request and someone else make the change manually.

klemens-morgenstern:
You pushed to your develop branch, that's fine.

PR #457, opened 2025-01-20
Fix several typos

striezel:
Fixes several typos, no functional changes intended.


klemens-morgenstern:
Something broke env access at some point for some reason. Gha is fun.

Windows is tested on drone, so this PR can ignore that

striezel:
> Something broke env access at some point for some reason. Gha is fun.

Thank you for the clarification.

PR #458, opened 2025-01-20
chore: remove executable flags from .cpp files

striezel:
The .cpp files should not have the executable flag set.

See <https://github.com/boostorg/admin/issues/47#issuecomment-2575165830>.


klemens-morgenstern:
Thanks

PR #459, opened 2025-01-23
Fix include directives from boost/process/posix/*.hpp having wrong paths

SimonMaracine:
I tried to use some other launcher on Linux, but compilation was failing due to the wrong include paths.

(There may be more wrong paths that I haven't noticed, so please feel free to reject this pull request and make the fixes as you wish. I could have opened up an issue instead, but felt this is fine too.)


SimonMaracine:
Thank you!

PR #465, opened 2025-02-27
Issue/453

klemens-morgenstern:
@barracuda156 can you check if this works?

### DIFF
--- a/include/boost/process/v1/detail/posix/environment.hpp
+++ b/include/boost/process/v1/detail/posix/environment.hpp
@@ -107,7 +107,7 @@ public:
     using string_type = std::basic_string<char_type>;
     using native_handle_type = char_type **;
 
-    void reload() {this->_env_impl = ::environ;}
+    void reload() {this->_env_impl = environ;}
 
     string_type get(const pointer_type id) { return getenv(id); }
     void        set(const pointer_type id, const pointer_type value)
@@ -136,7 +136,7 @@ public:
     native_environment_impl & operator=(native_environment_impl && ) = default;
     native_handle_type _env_impl = environ;
 
-    native_handle_type native_handle() const {return ::environ;}
+    native_handle_type native_handle() const {return environ;}
 };
 
 
--- a/include/boost/process/v1/detail/posix/executor.hpp
+++ b/include/boost/process/v1/detail/posix/executor.hpp
@@ -275,7 +275,7 @@ class executor
         prepare_cmd_style_fn = exe;
         if ((prepare_cmd_style_fn.find('/') == std::string::npos) && ::access(prepare_cmd_style_fn.c_str(), X_OK))
         {
-            const auto * e = ::environ;
+            const auto * e = environ;
             while ((e != nullptr) && (*e != nullptr) && !boost::starts_with(*e, "PATH="))
                 e++;
 
@@ -316,7 +316,7 @@ public:
     const char * exe      = nullptr;
     char *const* cmd_line = nullptr;
     bool cmd_style = false;
-    char **env      = ::environ;
+    char **env      = environ;
     pid_t pid = -1;
     std::shared_ptr<std::atomic<int>> exit_status = std::make_shared<std::atomic<int>>(still_active);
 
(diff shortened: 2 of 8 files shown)


barracuda156:
@glebm Just in case, do you still have some machine to run the build of this? MacPorts port is broken for more reasons than this issue, but the one from my fork https://github.com/macos-powerpc/powerpc-ports/tree/dbcaaf046f38b15925ca8fedeec563e731fb029f/devel/boost187 is supposed to build. Patch updated in https://github.com/macos-powerpc/powerpc-ports/commit/dbcaaf046f38b15925ca8fedeec563e731fb029f for testing.

I also asked a few ppl on MacRumors, but no idea if someone responds on time.

P. S. Something may be broken on < 10.6, since I build everything on 10.6.8, as you know. But now the point is only to confirm that `environ` patch works (`dirfd` is identical to what I already used, so that one is fine).

glebm:
Hi @barracuda156, I'm traveling until Apr 19 with no access to my iBook. /cc @AJenbo

AJenbo:
Sorry can you fill me in on what I need to test? Should I build this branch (where can i find the build instructions) or run something?

barracuda156:
> Sorry can you fill me in on what I need to test? Should I build this branch (where can i find the build instructions) or run something?

@AJenbo Thank you. Could you pick the attached source and try building it?
If you use standard MacPorts in its default prefix, a simple one-time hack will be just to replace the folder:
```
sudo rm -R /opt/local/var/macports/sources/rsync.macports.org/macports/release/tarballs/ports/devel/boost187
sudo cp -pR [path_to_extracted_folder]/boost187 /opt/local/var/macports/sources/rsync.macports.org/macports/release/tarballs/ports/devel/
```
(Do not rerun `port sync` in this case, otherwise port source will be overwritten with the default.)

If you use an overlay local repo, just place `boost187` in `devel` directory, run `port sync`, then build it.

P. S. This was not tested on Tiger, perhaps by anyone, so I apologize in advance in something else fails. But hopefully this works with gcc14 (and I guess all dependencies should build fine on 10.4).

[boost187.zip](https://github.com/user-attachments/files/19050801/boost187.zip)

barracuda156 (approved):
@klemens-morgenstern Let me go ahead and approve this, even though I cannot run the build on a legacy macOS right now (but we got a deadline by tomorrow), since:

1. The patches here are very close to what worked for me to fix the build for gcc14 locally (but more complete than my earlier ad hoc fix), i.e. https://github.com/macos-powerpc/powerpc-ports/commit/b333599f02db7ef2f6b54122efe6ec6741d089e4 – so I am pretty sure the solution is correct.
2. I have verified the build on Sonoma with clang, so it is safe.

I will verify this myself once I physically can, but I think it will be good to have it merged in the coming release now.

Zorg-Org:
Hi!  doctor_dog here from [over at MacRumors](https://forums.macrumors.com/threads/macports-development-for-powerpc-10-4-10-5-10-6-unofficial-invitation-to-cooperate.2363509/page-9?post=33776370#post-33776370) (one of the folks asked to test there).  I can confirm I was able to build this on 10.5.8 ppc32 and python 3.12 without issues.

barracuda156:
@AJenbo Install MacPorts (I assume it is installed, since you have built gcc).
Then run `sudo port -v sync`.
Extract boost187 archive I attached above and replace original boost187 folder with the newly extracted:
```
sudo rm -R /opt/local/var/macports/sources/rsync.macports.org/macports/release/tarballs/ports/devel/boost187
sudo cp -pR [path_to_extracted_folder]/boost187 /opt/local/var/macports/sources/rsync.macports.org/macports/release/tarballs/ports/devel/
```
Then run `sudo port -v -n -s install boost187`.

barracuda156:
10.4 does not have `sys_proc` header, as I recall. It should be an easily fixable issue (and unrelated to the one addressed here).
It will probably fail on the same spot on Tiger regardless of the arch. SDK headers are nearly always arch-agnostic.

klemens-morgenstern:
Right, I'll add the `MAC_OS_X_VERSION_MIN_REQUIRED` separately.

AJenbo:
> 10.4 does not have `sys/proc_info.h` header, as I recall. It should be an easily fixable issue (and unrelated to the one addressed here). It will probably fail on the same spot on Tiger regardless of the arch. SDK headers are nearly always arch-agnostic.

Confirmed same error on the 10.4 PowerPC

barracuda156:
@AJenbo Thank you. Yes, that was expected. The header is apparently related to `libproc`, and it needs kernel support, which is missing in 10.4.11 ppc.

There are some discussions:
https://github.com/mistydemeo/tigerbrew/issues/426
https://trac.macports.org/ticket/58160

What I have seen or done myself was just conditionally disabling related functionality. I do not know how important it is here; certainly, older versions of Boost worked on Tiger.

PR #487, opened 2025-05-03
Remove Ubuntu 20.04 builds from CI

striezel:
The Ubuntu 20.04 image on GitHub Actions has been unavailable since 2025-04-15. See <https://github.com/actions/runner-images/issues/11101> for more information on the deprecation and removal.

Therefore all build jobs that use the Ubuntu 20.04 runner image of GHA will fail and have to be replaced by newer images or have to move into Ubuntu 20.04 containers.


klemens-morgenstern:
Thank you

PR #520, opened 2025-10-06
Fix 'unused variable' warning

EelisVP:
If BOOST_PROCESS_V2_DISABLE_NOTIFY_FORK is defined, `ctx` is unused.


klemens-morgenstern:
Thanks!

PR #538, opened 2025-12-22
Fix `/dev/fd` not found error

### LINKED ISSUE #529: Problem with process creating (boost process v2) in limited linux systems
Hello!
I'm using boost::process::v2 (boost version 1.89 - latest stable) in my project with cross-compiling the final binary. If I use the amd64 standard debian 12, all is ok and process creating ended successfully without errors. But if I using it in limited linux (system with busybox, specific process architecture (for example, armv7l)) I getting an unknown exception. Try/catch can't process that exception. Google breakpad also can't give me details.
I checked the system - clone(), execve(), fork() works there. But bp::v2::process give me an unknown exception.
Also I testing on aarch64 (arm64) limited system and on standard linux version system. In standard linux that works great, but on limited system I've got the same error.
**Warning**: the "limited" version I mean Linux core + standard utilities such as busybox, libc, sh and other configurated extremely necessary utils.
I tested work directory access - all is ok on my side. I running the "/bin/echo" manually from shell, and that works good. 

System info:
armv7l, linux version 4.19.91,
GNU C Library (Buildroot) stable release version 2.29.
BusyBox v1.24.1 (2024-03-14 16:36:28 CST) multi-call binary

Binary compiled with GCC 8.3.0

Error log:

```shell
Creating process...
terminate called after throwing an instance of 'boost::wrapexcept<boost::system::system_error>'
terminate called recursively
Aborted (core dumped)
``` 

Code:

```c++
#include <boost/process/v2/process.hpp>
#include <boost/process/v2/stdio.hpp>
#include <boost/program_options/errors.hpp>

#include <iostream>

int main()
{
    try {
        std::cout << "Creating process..." << std::endl;
        boost::process::v2::process proc(context, "/bin/echo", {"hello", "world!"}, boost::process::v2::process_stdio {});
        std::cout << "Process was created." << std::endl;
        context.run();

        return EXIT_SUCCESS;

    } catch (const boost::program_options::error& ex) {
        std::cout << "Bad CLI: " << ex.what() << std::endl;
    } catch (const boost::system::system_error& error) {
        std::cout << "Exception: " << error.what() << ". Error code: " << error.code() << std::endl;
    } catch (...) {
        std::cout << "Unknown exception has occured." << std::endl;
    }

    return EXIT_FAILURE;
}

```

AJIOB:
Fixes #529 

For now, kept an error if both `/dev/fd` and `/proc/self/fd` dirs are not found.

### DIFF
--- a/include/boost/process/v1/detail/posix/handles.hpp
+++ b/include/boost/process/v1/detail/posix/handles.hpp
@@ -27,8 +27,12 @@ inline std::vector<native_handle_type> get_handles(std::error_code & ec)
     std::unique_ptr<DIR, void(*)(DIR*)> dir{::opendir("/dev/fd"), +[](DIR* p){::closedir(p);}};
     if (!dir)
     {
-        ec = ::boost::process::v1::detail::get_last_error();
-        return {};
+        dir = {::opendir("/proc/self/fd"), +[](DIR* p){::closedir(p);}};
+        if (!dir)
+        {
+            ec = ::boost::process::v1::detail::get_last_error();
+            return {};
+        }
     }
     else
         ec.clear();
@@ -60,7 +64,7 @@ inline std::vector<native_handle_type> get_handles()
 
     auto res = get_handles(ec);
     if (ec)
-        boost::process::v1::detail::throw_error(ec, "open_dir(\"/dev/fd\") failed");
+        boost::process::v1::detail::throw_error(ec, "open_dir(\"/dev/fd\") and open_dir(\"/proc/self/fd\") failed");
 
     return res;
 }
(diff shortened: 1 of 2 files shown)


AJIOB:
Hi @klemens-morgenstern , 

As I can see, all failures are not related to my changes

AJIOB:
Hi @klemens-morgenstern ,

Looks like we need to actualize CI. CI wrotes the next line:
```
Node 20 is being deprecated. This workflow is running with Node 24 by default. If you need to temporarily use Node 20, you can set the ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true environment variable. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
```

PR #539, opened 2025-12-24
Replace the dependency on asio with asio_core

Lastique:
Boost.Process doesn't use any advanced Boost.ASIO features and switching to asio_core removes the transitive dependencies on Boost.Context and Boost.DateTime.


Lastique:
Ping? Sadly, this wasn't merged before 1.91.0.

PR #555, opened 2026-05-05
Enable adding Ubuntu toolchain PPA by default

Flamefire:
The default in Boost.CI will be false but some jobs still rely on this, so keep it for now the default.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -8,7 +8,7 @@
 # As the yaml syntax for Drone CI is rather limited.
 #
 #
-globalenv={'B2_CI_VERSION': '1', 'B2_VARIANT': 'release'}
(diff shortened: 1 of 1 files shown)


klemens-morgenstern:
I have no idea what this does, can you give me the cliff notes?

Flamefire:
This was a recent Slack discussion: Basically:
Boost.CI adds the `ubuntu-toolchain-r/test` PPA by default.
This allows installing older or newer compilers than available on the base Ubuntu image

However this is bad for 2 reasons:
- System packages are preferred, so mostly another OS image should be used
- The PPA download takes time and may fail even when it isn't actually used.

The best way is to copy the updated one from https://github.com/boostorg/boost-ci

PR #559, opened 2026-06-16
`libprocstat` is no longer used on FreeBSD

samuelvenable:
Update Jamfile to reflect that

### DIFF
--- a/build/Jamfile
+++ b/build/Jamfile
@@ -48,7 +48,6 @@ lib user32 ;
 lib ws2_32 ;
 
 lib kvm ;
-lib procstat ;
 
 lib boost_process
    : process_sources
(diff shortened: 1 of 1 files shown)


klemens-morgenstern:
Does that break builds on older systems?

samuelvenable:
> Does that break builds on older systems?

It uses `libkvm` instead now to serve the same exact purposes. I did that so all BSD/SunOS platforms would have the same dependencies, for consistency.

samuelvenable:
Also, `libkvm` afaik has been a part of FreeBSD for much longer than `libproctat`. So I'd imagine that would improve compatibility with systems, both old and new, and not make that worse. :)

PR #563, opened 2026-07-22
Make basic_process default constructor =delete'd

### LINKED ISSUE #562: The process class is not default constructible as expected from documentation
Based on the current documentation and code comments (boost process v1.90) I was expecting the `boost::process::v2::process` class (aka `basic_process`) to be default contructible:
```cpp
/** An empty process is similar to a default constructed thread. It holds an empty
handle and is a place holder for a process that is to be launched later. */
basic_process() = default;
```
However when trying to do that I get a compiler error (also see https://godbolt.org/z/nE93jvPqo):
```
<source>:4:33: error: call to implicitly-deleted default constructor of 'boost::process::v2::process' (aka 'basic_process<>')
    4 |     boost::process::v2::process p;
      |                                 ^
/app/boost/include/boost/process/v2/process.hpp:69:3: note: explicitly defaulted function was implicitly deleted here
   69 |   basic_process() = default;
      |   ^
/app/boost/include/boost/process/v2/process.hpp:340:34: note: default constructor of 'basic_process<>' is implicitly deleted because field 'process_handle_' has no default constructor
  340 |   basic_process_handle<Executor> process_handle_;
      |                                  ^
1 error generated.
Compiler returned: 1
```

Is the process class supposed to be default-constructible or not? If not the default constructor should probably be `= delete` instead of defaulted and the comment removed.

sjoubert:
The basic_process defaulted default constructor fails to compiles and since it is *NOT* expected to be default constructible this marks it as such and removes the misleading comment

Fixes #562

### DIFF
--- a/doc/reference/process.adoc
+++ b/doc/reference/process.adoc
@@ -35,9 +35,7 @@ struct basic_process
     typedef basic_process<Executor1> other;
   };
 
-  /** An empty process is similar to a default constructed thread. It holds an empty
-  handle and is a place holder for a process that is to be launched later. */
-  basic_process() = default;
+  basic_process() = delete;
 
   basic_process(const basic_process&) = delete;
   basic_process& operator=(const basic_process&) = delete;
@@ -59,7 +57,7 @@ struct basic_process
       const filesystem::path& exe,
       std::initializer_list<string_view> args,
       Inits&&... inits);
-  
+
   // Construct a child from a property list and launch it using the default launcher..
   template<typename Args, typename ... Inits>
   explicit basic_process(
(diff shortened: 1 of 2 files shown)


klemens-morgenstern:
Are you telling me that a defaulted default constructor is not expected to be deleted?

sjoubert:
> Are you telling me that a defaulted default constructor is not expected to be deleted?

A defaulted constructor tells the compiler to automatically synthesize its definition when it needs one. In this case, it will fail to do so because a sub-object is not default constructible itself, but that could depend on the template parameter,... The semantic meaning conveyed by a defaulted implementation is that it is an available method, if you want to prevent or document non-usage explicitly that's what `= delete` is for.

Here in practice we could probably also just remove the default constructor. Since there are other constructors, omitting the no argument one will prevent it from being made available by the compiler implicitly. But deleting it explicitly makes is very clear to users, through the compiler error message, that it's non-availability is intentional (versus a "no matching constructor found, see the following 3 billions template constructors")

